PR #46, opened 2023-03-29
compose tunnel agent esbuild

royra:
- compose-tunnel-agent: esbuild


packages/cli/src/lib/docker-proxy-client.ts
```diff
@@ -12,7 +12,6 @@ const DOCKER_PROXY_PORT = 3000
 const baseDockerProxyService: ComposeService = {
   build: {
     context: DOCKER_PROXY_DIR,
```
Yshayy, line 14: We might need to add `target: 'production'`

packages/compose-tunnel-agent/build.mjs
```diff
@@ -0,0 +1,15 @@
+import * as esbuild from 'esbuild'
+
+await esbuild.build({
+  entryPoints: ['main.ts', 'index.ts'],
```
Yshayy, line 4: What does it mean that we have two entry points?

Yshayy (commented):
Looks good, have you verified it works in a clean environment? (production setup, out of working dir/workspace node_modules, standalone cli)

packages/cli/src/lib/docker-proxy-client.ts
royra: the default is the last builder

PR #77, opened 2023-05-07
Add environment variables for tunnel server urls

### LINKED ISSUE #57: Add environment variables for tunnel-server urls for usage in docker-compose
When doing service-service communication, Preevy can work well due to compose service discovery.
However, when using a frontend that utilize the exposed port for connecting to backend (can be common in dev environments), it's a bit tricky since we need to replace the url with tunneling url which we don't know about yet.

One way to solve it, is by injecting/resolving env variables of tunneling url in compose normalization. (so it can affect both build args, and environment)

For example:
```yaml
services:
     api:
           ....
           ports:
           - 9005:3000
     fronend:
          environment:
          - API_URL=http://localhost:9005
```
This could be a common setup.

With Preevy, it could be something like:
```yaml
services:
     api:
           ....
           ports:
           - 9005:3000
     fronend:
          environment:
          - API_URL={PREEVY_BASE_URI_API_3000:-http://localhost:9005}
 ```
In the future, we can enhance this by supporting inference of this configuration automatically and replacing all static `http://localhost:xx` variables' values based on public ports definitions.

### PR DESCRIPTION
**Pull request by naorzr**
Expose tunneling URL as environment variable in the following template
```PREEVY_BASE_URI_SERVICENAME_SERVICEPORT```
- **Example**: PREEVY_BASE_URI_BACKEND_9006

closes #57

### CODE REVIEW

File: packages/cli/src/lib/commands/up/index.ts [hunk 1]
```diff
@@ -92,6 +96,15 @@ const up = async ({
   })
   const projectName = userSpecifiedProjectName ?? userModel.name
   const remoteDir = remoteProjectDir(projectName)
+
+  const exposedServices = getExposedTcpServices(userModel)
```
royra, line 100: it would probably simplify things if `getExposedTcpServices` returned just `[serviceName, portNumber]`, then you could do something like
```typescript
const env = Object.fromEntries(
  getExposedTcpServices(userMode)
  .map(([service, port]) => [
    `PREEVY_BASE_URI_${service.toUpperCase()}_${port}`, 
    tunnelUrl(...),
  ])
)
```

packages/cli/src/lib/compose/client.ts
```diff
@@ -4,6 +4,13 @@ import { WriteStream } from 'fs'
 import { ComposeModel } from './model'
 import { childProcessPromise, childProcessStdoutPromise } from '../child-process'
 
+export const getExposedTcpServices = (model: ComposeModel) => Object.entries(model.services ?? [])
+  .filter(x => x[1].ports)
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
royra, line 9: passing a type guard to the filter above will remove the null assertion.

packages/cli/src/lib/compose/client.ts
```diff
@@ -23,7 +30,7 @@ const composeClient = (
     stdin: Buffer.isBuffer(composeFiles) ? composeFiles : undefined,
   })
 
-  const getModel = async () => yaml.parse(await execComposeCommand(['convert'])) as ComposeModel
+  const getModel = async () => yaml.parse(await execComposeCommand(['convert', '--no-interpolate'])) as ComposeModel
```
royra, line 33: worth explaining with a comment why we need the `no-interpolate` arg

packages/cli/src/lib/tunneling/index.ts
```diff
@@ -25,17 +25,17 @@ export const flattenTunnels = (
 
 export class UnverifiedHostKeyError extends Error {
   constructor(
-      readonly tunnelOpts: TunnelOpts,
-      readonly hostKeySignature: string,
+    readonly tunnelOpts: TunnelOpts,
```
royra, line 28: can you undo formatting-only changes in unrelated code? better to turn off prettify for this project if you have it on.

packages/cli/src/lib/tunneling/index.ts
```diff
@@ -111,3 +111,19 @@ export const ensureTunnelKeyPair = async (
   await store.setTunnelingKey(Buffer.from(keyPair.privateKey))
   return keyPair
 }
+
+interface Props {
```
royra, line 115: nit: prefer types over interfaces unless needed (e.g, recursive types, declaration merging). also - if it's only used once, define it inline.

packages/common/src/ssh/base-client.ts
```diff
@@ -5,6 +5,11 @@ import { tryParseJson } from '../json'
 import { Logger } from '../log'
 import { formatPublicKey, parseKey } from './keys'
 
+export interface BaseUrl {
```
royra, line 8: Not a huge difference, but I would use a URL type which is more extendable. what if there's a base path, e.g, https://livecycle.run/myorg?

packages/common/src/tunnel/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  name: string
+  project: string
+  port: number
+}) => {
+  port: number
+  tunnel: string
+}
+
+export const tunnelNameResolver = (
+  { userDefinedSuffix }: { userDefinedSuffix?: string },
```
royra, line 14: remove `userDefinedSuffix` if it's not used

site/docs/recipes/service-discovery.md
```diff
@@ -0,0 +1,47 @@
+---
+title: "Service discovery using environment variables"
+sidebar_position: 1
+date: "2023-05-07"
+---
+
+## Problem
+
+In development environments, it's common for frontend applications to communicate with backend services using exposed ports. While Preevy handles service-service communication well with compose service discovery, it can be challenging when frontend applications need to connect to backend services via exposed ports. The tunneling URL must be replaced, but it's not known in advance.
```
royra, line 9: I don't think "compose service discovery" is a familiar term.

Maybe add explicit explanation that code running in the containers can access other containers using the name of the service as a hostname (https://docs.docker.com/compose/networking/), but code running in the browser does not.

site/docs/recipes/service-discovery.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Solution
+
+Preevy offers a simple solution for this problem by exposing the tunneling URL as an environment variable. The environment variable is named after the service name + port, with the prefix `PREEVY_BASE_URI`. For example, if the service name is `frontend` and is exposed on port 4000, the environment variable will be `PREEVY_BASE_URI_FRONTEND_4000`.
```
royra, line 13: Worth mentioning this is at compose build time rather than runtime.

royra (approved):
LGTM! minor comments

PR #80, opened 2023-05-11
Fix issue with injected env vars

AssafKr:


packages/cli/src/lib/compose/client.ts
```diff
@@ -54,7 +50,9 @@ export type ComposeClient = ReturnType<typeof composeClient>
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type ParametersExceptFirst<F> = F extends (arg0: any, ...rest: infer R) => any ? R : never;
 
-export const localComposeClient = (composeFiles: string[] | Buffer, projectName?: string) => {
+type LocalComposeClientArgs = {composeFiles: string[] | Buffer; projectName?: string; env?: NodeJS.ProcessEnv}
```
royra, line 53: nit: the existing convention in the code base is that a separate type is not defined unless it's used somewhere outside of the func definition

site/docs/recipes/service-discovery.md
```diff
@@ -4,6 +4,21 @@ sidebar_position: 1
 date: "2023-05-07"
 ---
 
+## tl;dr
```
royra, line 7:
```suggestion
## TL;DR
```

site/docs/recipes/service-discovery.md
```diff
@@ -4,6 +4,21 @@ sidebar_position: 1
...
+```yaml
+services:
+     service_name:
+          ...
+          ports: exposed_port:internal_port
+```
+
+Preevy will generate the following environment variable:
+
+`PREEVY_BASE_URI_{{ comopse_service_name }}_{{ exposed_port }}`
```
royra, line 20: Here is a shorter TL;DR, see if it works for you:

If the external URL is needed in the compose file (e.g, a BASE_URL environment variable), preevy will supply an environment variable which you can [interpolate](https://docs.docker.com/compose/compose-file/12-interpolation/) as follows:

```yaml
services:
     my_service:
          ...
          env:
            BASE_URL=${PREEVY_BASE_URI_MY_SERVICE_8000}
          ports:
            - 8000
```

royra (approved):
See comments

PR #83, opened 2023-05-14
extract core and driver packages

royra:


Yshayy:
I think we need to add the new packages to our GH actions, right?

Yshayy:
Also, add the information of using Lerna in our getting started. 
Are we going to use Lerna for publishing as well?

packages/core/src/compose/client.ts
```diff
@@ -36,8 +38,13 @@ const composeClient = (
   const getModel = async () => yaml.parse(await execComposeCommand(['convert'])) as ComposeModel
 
   return {
-    startService: (...services: string[]) => execComposeCommand(['start', ...services]),
     getModel,
+    getModelOrUndefined: async () => getModel().catch(e => {
```
Yshayy, line 42: Is that a fix for the missing compose file?

Yshayy: Also, currently I don't see any usage of that.

packages/cli/src/lib/machine/drivers/fake/index.ts
```diff
@@ -1,107 +0,0 @@
-import { Flags, Interfaces } from '@oclif/core'
-import { asyncMap } from 'iter-tools-es'
-import { SSHKeyConfig } from '../../../ssh/keypair'
-import { Machine, MachineDriver } from '../../driver'
-import { MachineCreationDriver, MachineCreationDriverFactory, MachineDriverFactory } from '../../driver/driver'
-
-const fakeMachine: Machine = {
```
Yshayy, line 7: I think maybe we should still have fake driver in core. (since it doesn't have dependency, it should be a problem)

Yshayy (commented):
Review almost done (around 70%).
Added some notes

Yshayy (commented):
The driver dependencies are much more nicely scoped.

Yshayy (commented):
Need also to bump version

packages/core/src/commands/up/index.ts
```diff
@@ -127,8 +126,6 @@ const up = async ({
     createCopiedFile('tunnel_server_public_key', formatPublicKey(hostKey)),
   ])
```
Yshayy, line 128: Why remove this information? It can be useful

royra:
> Also, add the information of using Lerna in our getting started. 
Not sure where to add it - for now it's only used to run `yarn build` in order (and parallelize the run of build and lint), so it's an implementation detail.

> Are we going to use Lerna for publishing as well?
Probably in the future, but now right now.

royra:
> Need also to bump version

This is not a user-visible feature - I was planning to release it with the GH plugin.

packages/core/src/compose/client.ts
royra: It gets used by the incoming plugins code.

packages/core/src/commands/up/index.ts
royra: it moved to line 97 together with the definition of `envId`

packages/cli/src/lib/machine/drivers/fake/index.ts
royra: Added the code, but I'm not sure it's useful. Its only purpose was to demo multiple drivers when we only had lightsail.

royra: Strike that - adding it brings back `@oclif/core` because of the flags definitions

PR #86, opened 2023-05-16
Microsoft Azure driver

eliyabar:
driver added, no snapshot functionality or profile storage to Azure.


packages/cli/src/lib/commands/up/machine.ts
```diff
@@ -101,12 +101,12 @@ export const ensureCustomizedMachine = async ({
       log.debug('Executing machine scripts')
       for (const script of machineDriver.customizationScripts ?? []) {
         // eslint-disable-next-line no-await-in-loop
-        await sshClient.execScript(script)
+        await sshClient.execScript(script, {})
       }
 
-      // ensure docker is accessible
+      log.info('Ensure docker is accessible')
```
royra, line 107: Not sure this implementation detail should be always printed.

phrasing
```suggestion
      log.info('Ensuring docker is accessible')
```

packages/cli/src/lib/machine/drivers/azure/client.ts
```diff
@@ -0,0 +1,242 @@
...
+  vm: VirtualMachine
+  publicIPAddress: string
+}
+
+export const nameGenerator = (alias: string) => (name: string, prefix = 'preevy') => `${`${prefix}${alias.replaceAll('-', '')}${name}`.toLowerCase()
+  .substring(0, 16)}${randomBytes(16)
+  .toString('hex')
+  .substring(0, 8)}`
+
+export const getVmName = (envId: string) => `vm${envId.replaceAll('-', '')}`
```
royra, line 66: I would go with a list of accepted characters, this way it won't break if `envId` contains more than `-` chars.

```typescript
.replace(/[^a-zA-Z0-9]/g, '')
```

packages/cli/src/lib/machine/drivers/azure/client.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export const getResourceGroupName = (envId: string) => `preevy-${envId}`
+
+function extractResourceNameFromId(rId: string) {
```
royra, line 70: can't use tags instead of storing stuff in the id?

packages/cli/src/lib/machine/drivers/azure/client.ts
```diff
@@ -0,0 +1,242 @@
...
+      await resourceClient.resourceGroups.beginDelete(resourceGroupName, {
+        forceDeletionTypes: 'Microsoft.Compute/virtualMachines,Microsoft.Compute/virtualMachineScaleSets',
+      })
+    },
+    listInstances: (): AsyncIterableIterator<VMInstance> =>
+      asyncMap(
+        async vm => (
+          { vm, ...(await getIpAddresses(networkClient, vm)) }),
+        asyncFilter<VirtualMachine>(({ tags }) =>
+          tags?.profile === profileId, computeClient.virtualMachines.listAll())
```
royra, line 141: a tag named `profile` is too generic, maybe `preevy-profile`?

packages/cli/src/lib/machine/drivers/azure/client.ts
```diff
@@ -0,0 +1,242 @@
...
+      ),
+
+    getInstance: async (envId: string): Promise<VMInstance> => {
+      const name = getVmName(envId)
```
royra, line 145: `envId` is not unique per subscription, there can be multiple profiles.

packages/cli/src/lib/machine/drivers/azure/index.ts
```diff
@@ -0,0 +1,231 @@
...
+const flags = {
+  region: Flags.string({
+    description: 'Microsoft Azure region in which resources will be provisioned',
+    required: true,
+  }),
+  'subscription-id': Flags.string({
+    description: 'Microsoft Azure subscription id',
+    required: true,
+  }),
+  'resource-group-name': Flags.string({
```
royra, line 99: shouldn't this be only in machine *creation*?

packages/cli/src/lib/machine/drivers/azure/index.ts
```diff
@@ -0,0 +1,231 @@
...
+          return undefined
+        }
+        throw e
+      })
+      if (!vmInstance) {
+        return undefined
+      }
+      return ({
+        ...machineFromInstance(vmInstance),
+        specDiff: [],
```
royra, line 193: looks like there's a missing implementation here

packages/cli/src/lib/machine/drivers/azure/types.ts
```diff
@@ -0,0 +1,15 @@
+type RootObjectDetailsError = {
```
royra, line 1: a file named `types` is an anti-pattern - just like you don't have files named `objects` or `classes`. put this code together with non-types where it belongs feature-wise.

packages/cli/src/lib/machine/drivers/azure/vm-creation-utils.ts
```diff
@@ -0,0 +1,255 @@
...
+    location,
+    tags,
+  }
+  return resourceClient.resourceGroups.createOrUpdate(
+    resourceGroupName,
+    groupParameters
+  )
+}
+
+export async function createStorageAccount(
```
royra, line 40: nit: function -> arrow

packages/cli/src/lib/machine/drivers/azure/vm-creation-utils.ts
```diff
@@ -0,0 +1,255 @@
...
+  return computeClient.virtualMachineImages.list(
+    location,
+    imageRef.publisher,
+    imageRef.offer,
+    imageRef.sku
+  )
+    .then((images: Array<VirtualMachineImageResource>) => images.find(x => x.name))
+}
+
+export async function getNICInfo(
```
royra, line 197: nit: `getNICInfo` -> `getNicInfo`

royra (commented):
see comments + needs rebase

packages/cli/src/commands/down/index.ts
```diff
@@ -52,7 +52,7 @@ export default class Down extends DriverCommand<typeof Down> {
     }
 
     await withSpinner(async () => {
-      await driver.removeMachine(machine.providerId, flags.wait)
+      await driver.removeMachine(machine.providerId, flags.wait, envId)
```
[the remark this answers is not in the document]
eliyabar: the envId sets the resource group name which I need to delete.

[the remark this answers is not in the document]
royra: you can encode whatever you want in the `providerId`, it can even contain multiple values using JSON.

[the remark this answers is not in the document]
eliyabar: wdym? id is a string. 
```
providerId = JSON.stringify({propA: ..., propB:...})
```

packages/cli/src/lib/machine/drivers/azure/client.ts
eliyabar: Microsoft uses the same pattern for all of its resource IDs. I can use tags, but they are mutable, whereas the ID isn't.

packages/cli/src/commands/ls.ts
```diff
@@ -19,7 +19,12 @@ export default class Ls extends DriverCommand<typeof Ls> {
   async run(): Promise<unknown> {
     const { flags } = await this.parse(Ls)
     const driver = await this.driver()
-    const machines = await asyncToArray(await commands.ls({ machineDriver: driver, log: this.logger }))
+    const machines = await asyncToArray(asyncMap(x => {
+      if ('publicIPAddress' in x) {
```
royra, line 23: can probably benefit from defining
```typescript
const isFullMachine = (o: PartialMachine): o is Machine => typeof (o as Machine).publicIPAddress === 'string'
```

then
```typescript
return { ...x, partial: !isFullMachine(x) }
```

and I think it can be used elsewhere as well.

packages/driver-azure/src/driver/client.ts
```diff
@@ -0,0 +1,238 @@
...
+        region,
+        ipConfigName,
+        resourceGroupName,
+        securityGroup,
+        networkClient
+      )
+      const vmImageInfo = await findVMImage(region, imageRef, computeClient)
+
+      if (!nic.id || !vmImageInfo?.name || !publicIPInfo.name || !nic.ipConfigurations?.[0].privateIPAddress) {
+        throw new Error('Could not create vm, missing properties')
```
royra, line 221: maybe you can include some of the checked props in the message so it will be easier to debug

packages/driver-azure/src/driver/vm-creation-utils.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+  const resourceGroupName = extractResourceGroupNameFromId(vm.id)
+  const nicName = extractResourceNameFromId(vm.networkProfile.networkInterfaces[0].id)
+  const nic = await getNicInfo(resourceGroupName, nicName, networkClient)
+  if (!nic.ipConfigurations?.[0].publicIPAddress?.id) {
+    throw new Error('publicIPAddress configuration not found')
+  }
+  const publicIPName = extractResourceNameFromId(nic.ipConfigurations[0].publicIPAddress.id)
+  const publicIPAddress = await networkClient.publicIPAddresses.get(resourceGroupName, publicIPName)
+  if (!publicIPAddress.ipAddress || !nic.ipConfigurations?.[0].privateIPAddress) {
```
royra, line 274: consider using `ensureDefined`/`extractDefined` (it's in the aws driver but it can be moved to core)

packages/driver-azure/src/driver/index.ts
```diff
@@ -0,0 +1,272 @@
...
+    subscriptionId,
+    profileId,
+  })
+
+  return {
+    customizationScripts: CUSTOMIZE_BARE_MACHINE,
+    friendlyName: 'Microsoft Azure',
+    getMachine: async ({ envId }) => cl.getInstance(envId).then(vm => machineFromInstance(vm)),
+    listMachines: async () => {
+      const vmInstances = await asyncToArray(cl.listInstances())
```
royra, line 95: this iterates over all the instances, maybe it's better to use `getInstance` per RG?

packages/driver-azure/src/driver/index.ts
```diff
@@ -0,0 +1,272 @@
...
+      const allResourceGroups = cl.listResourceGroups()
+      return asyncMap(async resourceGroup => {
+        const vm = typeof resourceGroup.name === 'string' && vmInstances.find(i => extractResourceGroupNameFromId(i.vm.id as string) === (resourceGroup.name as string).toLowerCase())
+        if (vm) {
+          return { ...machineFromInstance(vm),
+            partial: false } as (Machine|PartialMachine) & { envId: string }
```
royra, line 101: why is `partial` needed?

packages/core/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "declaration": true,
```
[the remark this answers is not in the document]
eliyabar: I actually not sure how types in TS work without it. I already added it to the rest of the services.

`Generate .d.ts files for every TypeScript or JavaScript file inside your project. These .d.ts files are type definition files which describe the external API of your module. With .d.ts files, tools like TypeScript can provide intellisense and accurate types for un-typed code.`

packages/driver-azure/src/driver/client.ts
```diff
@@ -0,0 +1,238 @@
...
+} from './vm-creation-utils'
+
+// Uncomment to see Azure logs (import @azure/logger)
+// setLogLevel('info')
+
+export class AzureResourceNotFound extends Error {
+  statusCode = 404
+  constructor(message: string) {
+    super(`AzureResourceNotFound ${message}`)
+    Object.setPrototypeOf(this, AzureResourceNotFound.prototype)
```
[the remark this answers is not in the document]
eliyabar: Actually copied it from their code, might be redundant.

packages/driver-azure/src/driver/vm-creation-utils.ts
```diff
@@ -0,0 +1,281 @@
...
+import {
+  ComputeManagementClient,
+  ImageReference,
+  Resource,
+  VirtualMachine,
+  VirtualMachineImageResource,
+} from '@azure/arm-compute'
+import { StorageManagementClient } from '@azure/arm-storage'
+
+export enum AzureCustomTags {
```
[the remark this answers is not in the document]
eliyabar: np, they are the same after all.

packages/driver-azure/src/driver/client.ts
```diff
@@ -0,0 +1,238 @@
...
+      const filter = `tagName eq '${AzureCustomTags.ENV_ID}' and tagValue eq '${envId}'`
+      const vmResource = await asyncFirst(asyncFilter(x => x.type === 'Microsoft.Compute/virtualMachines', resourceClient.resources.list({ filter })))
+      if (!vmResource?.id) {
+        throw new AzureResourceNotFound(`envId: ${envId}`)
+      }
+      const name = extractResourceNameFromId(vmResource.id)
+      const resourceGroupName = extractResourceGroupNameFromId(vmResource.id)
+      const vm = await computeClient.virtualMachines.get(resourceGroupName, name)
+      if (vm.provisioningState === 'Deleting') {
+        throw new AzureResourceNotFound(`current vm instance for envId: ${envId} is deleting`)
```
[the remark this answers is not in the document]
eliyabar: error, the wrapping func is looking if it's 404 and returns unknown if needed.
this is an edge case since we still have a VM but it is being deleted.

packages/driver-azure/src/driver/client.ts
eliyabar: sure

packages/driver-azure/src/driver/index.ts
eliyabar: it's a list that contains a machine or partial machine.
it's being filtered by type more easily.
what do you suggest? a machine with optional props?

packages/driver-azure/src/driver/index.ts
eliyabar: done

packages/driver-azure/src/driver/index.ts
royra: correct me if I'm wrong - deducing whether it's partial is just a matter of checking for other props, so the `partial` prop doesn't add new information and can just cause mismatching bugs.

PR #89, opened 2023-05-21
plugins

royra:


packages/cli/src/commands/up.ts
```diff
@@ -106,6 +107,7 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
       debug: flags.debug,
       machineDriver: driver,
       machineCreationDriver,
+      userModel: await this.ensureUserModel(),
```
Yshayy, line 110: The (this.ensureUserModel) can extracted to a variable?
We're using it also afterwards

packages/cli/src/commands/up.ts
```diff
@@ -119,14 +121,23 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
 
     const flatTunnels = flattenTunnels(tunnels)
 
+    const result = await asyncReduce(
+      { urls: flatTunnels },
+      async ({ urls }, envCreated) => envCreated(
+        { log: this.logger, userModel: await this.ensureUserModel() },
```
Yshayy, line 127: ensureUserModel() is used on every iteration, is that the right behavior?

packages/plugin-github-pr-link/src/commands/link-github-pr.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw new Error('No GitHub config provided for plugin - specify env vars, flags or populate the compose file')
+    }
+
+    const log = this.logger
+    const projectName = (await this.ensureUserModel()).name
+    log.debug(`project: ${projectName}`)
+    const envId = this.flags.id || await findAmbientEnvId(projectName)
+    log.debug(`envId: ${envId}`)
+
+    const urls = await this.config.runCommand('urls', [
```
Yshayy, line 41: I'm a bit confused.
This plugin can be used both as a hook and as a command?
In the command case, the logic is a bit weird as it uses the urls, couldn't it enable the plugin hooks for this run.

packages/cli/src/commands/up.ts
royra: it's cached (but fixed anyway)

packages/plugin-github-pr-link/src/commands/link-github-pr.ts
royra: Plugins can add commands and flags.
This command is separate from `up` and doesn't include its logic - it just upserts the GitHub comment.

packages/cli/src/commands/up.ts
royra: done

packages/cli-common/src/hooks/init/load-plugins.ts
```diff
@@ -0,0 +1,88 @@
...
+  (acc, { hooks }) => {
+    Object.entries(hooks || {})
+      .forEach(([name, fn]) => { (acc[name as HookName] ||= []).push(fn as HookFunc<HookName>) })
+    return acc
+  },
+  {} as Record<HookName, HookFunc<HookName>[]>,
+)
+
+export const initHook: OclifHook<'init'> = async function hook({ config, id: _id, argv }) {
+  const { flags } = await new Parser({
```
Yshayy, line 37: Wouldn't this cause slow down for all commands that are not dependent on compose. (like version, help or any profile command)

packages/cli/package.json
```diff
@@ -51,6 +50,7 @@
     "oclif": "^3",
     "shx": "^0.3.3",
```
Yshayy, line 51: Does the CLI need to have depednecy on the plugin for it to work?
(assuming our first plugins would be pre-bundled)

packages/cli-common/src/hooks/init/load-plugins.ts
```diff
@@ -0,0 +1,88 @@
...
+  ))
+
+  const initResultsWithConfigs = await Promise.all(
+    plugins.map(async p => ({
+      initResults: await p.plugin.init({ userModel, oclifConfig: config, pluginConfig: p.config, argv }),
+      config: p.config,
+    })),
+  )
+
+  const pluginCommands = initResultsWithConfigs
```
Yshayy, line 65: The internals are very complex, might worth adding some documentation or extracting logic to some functions.

It might worth limiting plugin.init to sync ops only to prevent waiting.

The loadCommands seems very internal, do we know what the effects of using it?

Yshayy (approved):
This change is big, but looks good overall!

Added some notes.
Two points worth addressing:
- How can we add plugin usage automatically in our GH action
- Init performane, cost of plugins init, loading oclif commands (also for commands that doesn't use compose at all)

If the cost is big, I'm not sure it worth the complexity and performance compared to adding support for dynamic flags with a prefix.

packages/plugin-github-pr-link/src/commands/link-github-pr.ts
Yshayy: So, it's to set an example?

Yshayy: Or its worth keeping for testing?

Yshayy: I worry that it might cause confusion

Yshayy (changes requested):
Also need to address the bundling/dep issue to make sure it works in production

packages/cli-common/src/hooks/init/load-plugins.ts
royra: yes, of about 0.5s.

but I guess that's the price of having the compose file as a source of configuration? I imagine it wouldn't be just for plugins.

I guess we can exclude those commands, but that would mean they wouldn't support plugins/configuration.

packages/plugin-github-pr-link/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+This plugin allows showing the Preevy environment URLs in a GitHub PR comment. The comment is added/updated at the `up` command and deleted in the `down` command.
+
+![Demo comment](./demo.png)
+
+## Configuration
+
+Add the plugin to the `plugins` section of the `x-preevy` element in your Docker Compose file:
+
+```yaml
```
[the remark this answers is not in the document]
royra: Not sure it's easy since the only thing currently configuring a plugin is the compose file.

We can add another compose file which has the plugin definition, but that would mean the "compose files" argument would become required, which is awkward.

I guess we need another way to indicate that a plugin is enabled. Maybe env var?

packages/cli/package.json
royra: If we want the user to not have to install the plugin, I guess we can specify it as an explicit dep.

BTW: if the plugin is not specified in the deps, not sure how it would work with npx.

packages/plugin-github-pr-link/README.md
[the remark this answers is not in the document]
Yshayy: Env var can work as well.
Basically, we already have mechanism for disabling the plugin, could be similar

packages/cli-common/src/hooks/init/load-plugins.ts
royra: It would be hard to sandbox the plugin from the core functionality, and I'm not sure it's worth it.
Plugins for other devx Node.js projects like eslint, webpack, jest, don't offer such protection and are vulnerable to exploits and supply chain attacks.
I don't think we should address this. IMO it's the user's responsibility to see if the plugins work for them, like any other package they add to their project.

packages/cli/package.json
royra: ok, added to the deps.

still need to solve the case of npx + unbundled plugins

packages/cli-common/src/hooks/init/load-plugins.ts
royra: refactored to extract the logic to named funcs in a `plugins` module in the `cli-common` package

packages/plugin-github-pr-link/README.md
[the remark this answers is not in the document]
Yshayy: I wan to avoid additional configuration file and flag at this point, why it's needed? 

Also, it doesn't solve the GH action issue, it makes it more complex compared to just setting env bar.

packages/cli/package.json
Yshayy: Why wouldn't it work with npx?

packages/cli-common/src/hooks/init/load-plugins.ts
Yshayy: The problem is not security or sandboxing, it's about creating a way to easily develop plugin and create contracts that aim for path of success. 
We need to choose the extension points and do it carefully.
0.5s is slow enough that we'll need to soon add another loader of loading plugins. (And that's without having actually list of plugins).

packages/cli/package.json
royra: I guess it can always work with `-p`:
```bash
npx -p preevy -p @bla/my-preevy-plugin ...
```

packages/plugin-github-pr-link/README.md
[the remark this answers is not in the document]
Yshayy: Got it, it support multiple flags and merging like compose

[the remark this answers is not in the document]
Yshayy: It can work well, and we'll need such file at some point, the only thing I'm a bit worried, is what if we'll want/need a different configuration format.
Enabling/disabling plugin with env var seems relatively simple.

Yshayy (approved):

packages/plugin-github-pr-link/README.md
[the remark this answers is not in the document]
royra: Env vars are a limited solution If you need to specify a configuration for the plugin.

Issue #68, opened 2023-05-02
Basic Authn/Authz support


Yshayy:
Initial implementation will be based on http basic authentication.
- Support for declaring "private" services
- Support for authentication using tunneling-key based authentication

PR #92, opened 2023-05-22
support private connection requests

naorzr:
as part of #68

### DIFF
--- a/tunnel-server/index.ts
+++ b/tunnel-server/index.ts
@@ -56,7 +56,7 @@ const sshServer = createSshServer({
     await envStore.set(key, { target: localSocketPath, clientId, publicKey, access })
     tunnelsGauge.inc({clientId})
   },
-  onPipeDestroyed: async (clientId, remotePath) => {
+  onPipeDestroyed: async ({clientId, remotePath}) => {
     const key = tunnelName(clientId, remotePath);
     sshLogger.debug('deleting tunnel %s', key)
     await envStore.delete(key)
--- a/tunnel-server/index.ts
+++ b/tunnel-server/index.ts
@@ -31,12 +31,7 @@ const BASE_URL = (() => {
 
 type BaseUrl = typeof BASE_URL
 
-const envStore = inMemoryPreviewEnvStore({
-  test: {
-    clientId: "",
-    target: 'http://3.73.126.120',
-  },
-})
+const envStore = inMemoryPreviewEnvStore()
 
 const logger = pino(appLoggerFromEnv())
 const app = createApp({ sshPublicKey, isProxyRequest: isProxyRequest(BASE_URL), proxyHandlers: proxyHandlers({envStore, logger}), logger })
@@ -55,10 +50,10 @@ const sshServer = createSshServer({
   log: sshLogger,
   sshPrivateKey,
   socketDir: '/tmp', // TODO
-  onPipeCreated: async (clientId, remotePath, localSocket) => {
+  onPipeCreated: async ({clientId, remotePath, localSocketPath, publicKey, access}) => {
     const key = tunnelName(clientId, remotePath);
-    sshLogger.debug('creating tunnel %s for localSocket %s', key, localSocket)
-    await envStore.set(key, { target: localSocket, clientId })
+    sshLogger.debug('creating tunnel %s for localSocket %s', key, localSocketPath)
+    await envStore.set(key, { target: localSocketPath, clientId, publicKey, access })
     tunnelsGauge.inc({clientId})
   },
   onPipeDestroyed: async (clientId, remotePath) => {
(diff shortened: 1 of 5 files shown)


tunnel-server/index.ts
```diff
@@ -55,10 +50,10 @@ const sshServer = createSshServer({
   log: sshLogger,
   sshPrivateKey,
   socketDir: '/tmp', // TODO
-  onPipeCreated: async (clientId, remotePath, localSocket) => {
+  onPipeCreated: async ({clientId, remotePath, localSocketPath, publicKey, access}) => {
```
Yshayy, line 53: I think that for consistency, both onPipeCreated and onPipeDestroyed should use props style signature

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
         return;
       }
 
+      if(env.access === 'private'){
+        const authorization = req.headers['authorization'];
+        if (!authorization) {
+          return unauthorized(res)
+        }
+
+        const basicAuth = Buffer.from(authorization?.split('Basic ')[1] || '', 'base64').toString('ascii');
```
Yshayy, line 67: Might worth splitting the authentication scheme and check if it's basic.
http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
...
+        const [username, password] = basicAuth.split(':');
+        if (username !== 'x-preevy-key'){
```
Yshayy, line 69: Can be `x-preevy-profile-key` to be more specific

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
...
+          return unauthorized(res)
+        }
+        const signature = Buffer.from(password, 'base64');
```
Yshayy, line 72: If the data is already wrapped in base64, is there any text format that can be stored binary data more efficiently and still passed in clear text and in url?

Yshayy (commented):
LGTM :)
Added few notes

tunnel-server/src/proxy.ts
naorzr: for arbitrary binary data base64 is the most space efficient

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
         return;
       }
 
+      if(env.access === 'private'){
```
royra, line 61: extract to function?

tunnel-server/src/ssh-server.ts
```diff
@@ -2,12 +2,26 @@ import crypto, { randomBytes } from 'crypto'
...
 import path from 'path'
-import ssh2 from 'ssh2'
+import ssh2, { ParsedKey, SocketBindInfo } from 'ssh2'
 import { inspect } from 'util'
 
 const idFromPublicSsh = (key: Buffer) =>
   crypto.createHash('sha1').update(key).digest('base64url').replace(/[_-]/g, '').slice(0, 8).toLowerCase()
 
+const getRequestedSocketPath = (info: ssh2.SocketBindInfo) => {
+  const [path, access] = info.socketPath.split('#')
```
royra, line 12: I would choose an extendable format like
`mypath;auth=profile-key`
this way we can add more stuff here

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
...
+        const authorization = req.headers['authorization'];
+        if (!(authorization && authorization.startsWith('Basic '))) {
+          return unauthorized(res)
+        }
+
+        const basicAuth = Buffer.from(authorization?.split('Basic ')[1] || '', 'base64').toString('ascii');
+        const [username, password] = basicAuth.split(':');
+        if (username !== 'x-preevy-profile-key'){
+          return unauthorized(res)
```
royra, line 70: add log

tunnel-server/src/proxy.ts
```diff
@@ -52,6 +58,24 @@ export function proxyHandlers({
...
 
+      if(env.access === 'private'){
+        const authorization = req.headers['authorization'];
+        if (!(authorization && authorization.startsWith('Basic '))) {
+          return unauthorized(res)
+        }
+
+        const basicAuth = Buffer.from(authorization?.split('Basic ')[1] || '', 'base64').toString('ascii');
+        const [username, password] = basicAuth.split(':');
+        if (username !== 'x-preevy-profile-key'){
```
royra, line 69: maybe the username should be the clientId?

PR #100, opened 2023-05-30
Private requests through compose tunnel agent

naorzr:


packages/compose-tunnel-agent/src/docker.ts
```diff
@@ -32,6 +33,7 @@ const client = ({
   ).map(x => ({
     project: x.Labels['com.docker.compose.project'],
     name: x.Labels['com.docker.compose.service'],
+    access: (x.Labels['preevy.access'] || 'public') as ('private' | 'public'),
```
royra, line 36: How would "saas-private" access work? isn't this another value of `preevy.access`?

packages/compose-tunnel-agent/src/ssh/tunnel-client.ts
```diff
@@ -131,26 +125,35 @@ export const sshClient = async ({
     clientId, tunnels: tunnelsFromHelloResponse(tunnels),
   })
 
+  const getForwardRequestId = (service: {name: string; project: string; port: number; access: 'private' | 'public'}) => {
```
royra, line 128: reconstructing the request ID from the details is code duplication, plus it can be ambiguous if more key-value pairs are introduced. instead the original ID needs to be saved with the forward object in `tunnelSet`.

packages/compose-tunnel-agent/src/ssh/tunnel-client.ts
```diff
@@ -131,26 +125,35 @@ export const sshClient = async ({
...
-      const forwards = new Map(
-        services.flatMap(service => service.ports.map(x => tunnelNameResolver({ ...service, port: x }))
-          .map(({ port, tunnel }) => ([tunnel, { host: service.name, port, service }])))
-      )
-
-      const inserts = maps.difference(forwards, existingForwards)
-      const deletes = maps.difference(existingForwards, forwards)
-
-      const haveChanges = inserts.size > 0 || deletes.size > 0
+      const forwards = tunnelSet()
```
royra, line 138: this code should be encapsulated into the tunnelSet "constructor"

packages/compose-tunnel-agent/src/ssh/tunnel-client.ts
```diff
@@ -131,26 +125,35 @@ export const sshClient = async ({
...
+      services.flatMap(service =>
+        service.ports.map(x => ({ forwardRequestId: getForwardRequestId({ ...service, port: x }), port: x }))
+          .forEach(({ port, forwardRequestId }) => forwards.add(
+            { host: service.name, port, service, sockets: new Set(), forwardRequestId }
+          )))
+
+      const deletes = existingForwards.difference(forwards)
```
royra, line 145: `inserts` and `deletes` can both be returned from `.difference`

packages/compose-tunnel-agent/src/ssh/tunnel-set.ts
```diff
@@ -0,0 +1,50 @@
+import net from 'net'
+import { RunningService } from '../docker'
+
+interface Forward {
```
royra, line 4: prefer type to interface

packages/compose-tunnel-agent/src/ssh/tunnel-set.ts
```diff
@@ -0,0 +1,50 @@
...
+  service: RunningService
+  host: string
+  port: number
+  sockets: Set<net.Socket>
+  forwardRequestId: string
+}
+
+const forwardKey = (forwardRef: string) => forwardRef.replace(/^\/|#.*/g, '')
+
+interface TunnelSet {
```
royra, line 14: is this interface needed? it's not exported

tunnel-server/src/ssh-server.ts
```diff
@@ -4,23 +4,34 @@ import net from 'net'
...
+const schema = z.object({
+  path: z.string(),
+  params: z.object({ access: z.enum(['public', 'private']).default('public') }).required(),
+})
+
+export const parseRequestedPath = (requestPath: string) => {
+  const [path, params] = requestPath.split('#')
+
+  if (!params) {
+    return schema.safeParse({ path, params: {} })
```
royra, line 21: not sure this is worth a schema validator, as most of the parsing is done via custom code anyways

tunnel-server/src/ssh-server.ts
```diff
@@ -86,12 +97,18 @@ export const sshServer = ({
         }
 
         if ((name as string) == 'cancel-streamlocal-forward@openssh.com') {
-          const requestedSocketPath = getRequestedSocketPath(info as unknown as SocketBindInfo).path
-          if (!tunnels.delete(requestedSocketPath)) {
-            log.error('cancel-streamlocal-forward@openssh.com: socketPath %j not found, rejecting', requestedSocketPath)
+          const res = getRequestedSocketPath(info as unknown as SocketBindInfo)
+          if(res.success === false){
```
royra, line 101: can success be anything but a boolean?

tunnel-server/src/ssh-server.ts
naorzr: the parsing applies defaults where's needed.

tunnel-server/src/ssh-server.ts
naorzr: nop

packages/compose-tunnel-agent/src/ssh/tunnel-set.ts
naorzr: yup, otherwise you can't refere to it in the difference function signature

packages/compose-tunnel-agent/src/ssh/tunnel-set.ts
naorzr: we had this in a previous PR.
there is no significant difference, unless you need specific features, and in that case if u have to choose, the typescript team actually recommends interface.

packages/compose-tunnel-agent/src/ssh/tunnel-client.ts
naorzr: there is an importence to doing them one after the other, because we first delete, and then check to see what needs to be inserted. wouldn't work otherwise..

packages/compose-tunnel-agent/src/docker.ts
naorzr: depends on what saas do u refer to.. but we can always add authentication type.. wether it's basic or other

packages/compose-tunnel-agent/src/ssh/tunnel-client.ts
naorzr: you should view the usage in the file..
we need to construct the requestId from running services.. no much other way around it

PR #104, opened 2023-06-05
Some telemetry improvements and fixes

Yshayy:
The major change is using device_id as the distinct_id instead of run_id, and using group identity to specify link device to profile.

In addition:
- Added devcontainer for easier testing/development
- Updated telemetry endpoint to preevy url
- Stack traces are now hidden unless using DEBUG variable. (https://oclif.io/docs/debugging)
- Add manual flush of telemetry of errors by adding catch handler to basic command.
- Some improvements to the throttle/flush logic, previously flush was executed 3 times:
```
Capture event run -> trigger first flush
Capture identity -> didn't trigger flush (throttle)
Post Run -> trigger flush (identity sent)
BeforeExit -> Capture exit event and flush
```
After this update, flush will happen only when:
- After capture event occurred without any sequential event in past 3s (with limited window of 8s)
- BeforeExit or Error handling
While there's an additional hook for `exit` event, it doesn't really manage to send these events. (unless we change it to blocking I/O or persist if the next run)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,12 +6,12 @@
 	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
 
 	"features": {
-		"ghcr.io/devcontainers/features/docker-in-docker:2": {
-			"version": "latest",
-			"enableNonRootDocker": "true",
-			"moby": "true"
-		},
-		"ghcr.io/devcontainers/features/node:1": {}
+		"ghcr.io/devcontainers/features/docker-in-docker:2": {},
+		"ghcr.io/devcontainers/features/node:1": {},
+		"ghcr.io/devcontainers/features/aws-cli:1": {},
+		"ghcr.io/devcontainers/features/azure-cli:1": {},
+		"ghcr.io/dhoeric/features/google-cloud-cli:1": {},
+		"ghcr.io/rio/features/k3d:1": {}
 	}
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.


packages/cli/bin/run
```diff
@@ -1,9 +1,8 @@
 #!/usr/bin/env node
-
 require('source-map-support').install()
 
 const oclif = require('@oclif/core')
 
-oclif.settings.debug = true
+oclif.settings.debug = !!process.env["DEBUG"]
```
royra, line 6: nit

```suggestion
oclif.settings.debug = Boolean(process.env.DEBUG)
```

packages/core/src/telemetry/emitter.ts
```diff
@@ -82,29 +88,46 @@ export const telemetryEmitter = async ({ dataDir, version, debug }: {
...
+      props = args[1] as TelemetryProperties
+    } else {
+      [groupIdentityType, groupId, props] = args as [GroupIdentityType, string, TelemetryProperties ]
+      groupIdentities[groupIdentityType] = groupId
+    }
 
-    if (isCurrentIdAnonymous || Object.keys(person ?? {}).length) {
+    if (groupIdentityType && groupId) {
+      pushEvent({
+        event: '$groupidentify',
```
royra, line 107: https://posthog.com/docs/getting-started/group-analytics#the-difference-between-groups-and-cohorts:

```
If your only goal is to create a list of users with something in common, we recommend cohorts instead of groups.
```

packages/core/src/telemetry/emitter.ts
```diff
@@ -123,9 +147,15 @@ export const telemetryEmitter = async ({ dataDir, version, debug }: {
     setProps: (props: TelemetryProperties) => {
       Object.assign(commonProperties, props)
     },
-    flush: () => {
-      throttledFlush.cancel()
-      return flush()
+    // For making sure we're not keeping the process running due to debounce.
+    // Flush is not called as there could be other captures afterwards such as exit event.
+    release: () => {
```
royra, line 152: I think a more idiomatic name is `unref` - see [timeout](https://nodejs.org/api/timers.html#timeoutunref), [worker thread](https://nodejs.org/api/worker_threads.html#workerunref), [socket](https://nodejs.org/api/dgram.html#socketunref) (there's more).

packages/core/src/telemetry/emitter.ts
```diff
@@ -113,8 +136,9 @@ export const telemetryEmitter = async ({ dataDir, version, debug }: {
       pushEvent({
         event,
         timestamp: new Date(),
-        distinct_id: currentId,
+        distinct_id: machineId,
```
royra, line 139: I think this change might disable [event linking](https://posthog.com/docs/getting-started/identify-users#linking-past-events-with-our-new-user). this is why it was using a mutable `currentId` which starts with an anonymous ID and later set to the profile at `identify`

packages/cli/src/profile-command.ts
```diff
@@ -55,15 +59,15 @@ abstract class ProfileCommand<T extends typeof Command> extends BaseCommand<T> {
...
+      this.error('Store was not initialized')
     }
     return this.#store
   }
 
   #profile: Profile | undefined
   get profile(): Profile {
     if (!this.#profile) {
-      throw new Error(`Profile not initialized, run ${chalk.italic.bold.greenBright('preevy init')} to get started.`)
+      this.error(`Profile not initialized, run ${chalk.italic.bold.greenBright('preevy init')} to get started.`)
```
royra, line 70: 👍🏽

packages/core/src/telemetry/emitter.ts
```diff
@@ -2,19 +2,21 @@ import os from 'os'
...
-const FLUSH_INTERVAL = 5000
+const TELEMETRY_URL = 'https://telemetry.preevy.dev/v1/event'
+
+export type GroupIdentityType = 'profile'
 
 type IdentifyFunction = {
-  (person: TelemetryProperties): void
-  (id: string, person?: TelemetryProperties): void
+  (props: TelemetryProperties): void
+  (identityType: GroupIdentityType, props: TelemetryProperties): void
```
royra, line 18: what's the use case of calling `identify` with `identifyType` but no `id`?

packages/core/src/telemetry/emitter.ts
Yshayy: A profile can also have several users, so by using deviceId and profile as group, it's simplify the process of understanding how each user in a profile interact with preevy. Currently, switching a profile is linking your identity and it's a bit weird. In addition, using a session ID as an identifier creates too much noise.

As for linking:
It basically means that there's no linking for now (unless in the future we'll have a user identifier instead of a device identifier).

packages/core/src/telemetry/emitter.ts
Yshayy: Using a group makes sense in it because we want to aggregate events based on profile.

PR #105, opened 2023-06-06
expose username and password through the cli

naorzr:
Updated by @Yshayy:
- Tunnel server expects jwt on basic auth instead of signed data. (we could add support for bearer in the future)
- CLI generate jwt when using `--include-access-credentials` flags
- Use EdDSA as default for ssh keys (https://goteleport.com/blog/comparing-ssh-keys/)

I also added Issuer claim and validation of that claim (preevy://thumbprint), not sure if it's needed for now, but the idea was to easily separate between other JWT to uses external issuer.

The auth part can be later expanded to session cookie, to support websocket and cross-environment browser session.


packages/core/src/credentials/index.ts
```diff
@@ -0,0 +1,19 @@
+import { parseKey } from '@preevy/common'
+import { memoize } from 'lodash'
+
+export const getUserCredentials = memoize((privateKey: string | Buffer) => {
+  const parsedKey = parseKey(privateKey)
+  const user = 'x-preevy-profile-key'
```
Yshayy, line 6: User can be replaced in favor of data that contains information about the password/token scope. (like timestamp, or jwt structure)

packages/core/src/commands/up/index.ts
```diff
@@ -168,8 +171,18 @@ const up = async ({
...
       })
-
-      return queryResult.tunnels
+      if (!includeAccessCredentials) {
+        return queryResult.tunnels
+      }
+
+      return queryResult.tunnels.map(x => ({
+        ...x,
+        ports: Object.fromEntries(
```
royra, line 180: mapValues?

royra (approved):
minor comments

packages/cli/src/commands/up.ts
```diff
@@ -96,16 +102,19 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
       {
         service: { header: 'Service' },
         port: { header: 'Port' },
         url: { header: 'URL' },
       },
-      this.flags,
+      {
+        ...this.flags,
+        'no-truncate': this.flags['no-truncate'] ?? (!this.flags.output && !this.flags.csv && flags['include-access-credentials']),
+        'no-header': this.flags['no-header'] ?? (!this.flags.output && !this.flags.csv && flags['include-access-credentials']),
```
royra, line 116: IMO this is redundant, let user decide what they want

packages/cli/src/commands/urls.ts
```diff
@@ -13,6 +13,10 @@ export default class Urls extends ProfileCommand<typeof Urls> {
     ...envIdFlags,
     ...tunnelServerFlags,
     ...ux.table.flags(),
+    'include-access-credentials': Flags.boolean({
```
royra, line 16: extract to existing cli-common/flags

packages/cli/src/commands/urls.ts
```diff
@@ -44,15 +48,18 @@ export default class Urls extends ProfileCommand<typeof Urls> {
...
       tunnelOpts,
       knownServerPublicKeys: pStore.knownServerPublicKeys,
-      tunnelingKey: await pStore.getTunnelingKey(),
+      tunnelingKey,
       log: this.logger,
     })
 
     const flatTunnels = await commands.urls({
       rootUrl,
+      includeAccessCredentials: !!flags['include-access-credentials'],
```
royra, line 61: `!!` seems redundant (if there's a default)

packages/core/src/credentials/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      return 'EdDSA'
+    default:
+      throw new Error('Only RSA and Ed25519 keys are supported')
+  }
+}
+
+export const jwtGenerator = (privateKey: string | Buffer) => {
+  const sshKey = ssh2.utils.parseKey(privateKey)
+  if (sshKey instanceof Error) {
+    throw new Error(`Could not parse private key: ${sshKey.message}`)
```
royra, line 23: can throw the returned `Error` (or at least set `cause` to it)

packages/core/src/commands/up/index.ts
```diff
@@ -173,7 +177,7 @@ const up = async ({
     const composeArgs = calcComposeArgs({ userSpecifiedServices, debug, cwd })
     log.debug('Running compose up with args: ', composeArgs)
     await withDockerSocket(() => compose.spawnPromise(composeArgs, { stdio: 'inherit' }))
-
+    const credentials = await getUserCredentials(jwtGenerator(sshTunnelPrivateKey))
```
royra, line 180: I would extract this out of the `withDockerSocket` callback since the docker socket is not needed here

packages/core/src/commands/up/index.ts
```diff
@@ -184,8 +188,16 @@ const up = async ({
...
       })
-
-      return queryResult.tunnels
+      if (!includeAccessCredentials) {
+        return queryResult.tunnels
+      }
+
+      return queryResult.tunnels.map(x => ({
+        ...x,
+        ports: mapValues(x.ports, urls => urls.map(url =>
```
royra, line 197: extract to function, e.g, `addBasicUserCredentialsToTunnels`

packages/core/src/commands/urls.ts
```diff
@@ -12,7 +16,11 @@ export const urls = async ({ envId, rootUrl, clientId, projectName, serviceAndPo
 
   const { tunnels } = await queryTunnels({ tunnelUrlForService, retryOpts: { retries: 2 } })
 
+  const credentials = includeAccessCredentials ? await getUserCredentials(jwtGenerator(tunnelingKey)) : undefined
   return flattenTunnels(tunnels)
+    .map(x => ({ ...x,
+      url: credentials
```
royra, line 22: this is a different version of the logic at up/index.ts

packages/core/src/credentials/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    user,
+    password: await jwtGen(),
+  }
+}
+
+export const addBasicAuthCredentials = (url: string, user: string, password: string) => {
+  const parsed = new URL(url)
+  parsed.username = user
+  parsed.password = password
+  return parsed.toString()
```
royra, line 51:
```suggestion
export const addBasicAuthCredentials = (url: string, user: string, password: string) => Object.assign(new URL(url),  {
  username: user,
  password,
}).toString()
```

tunnel-server/src/auth.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+function extractBasicAuth(req: IncomingMessage){
+  const authorization = req.headers['authorization'];
+  const [scheme, data] = authorization?.split(' ') ?? []
+  if (scheme !== 'Basic') {
+    return undefined
+  }
+  const basicAuth = Buffer.from(data, 'base64').toString('ascii');
+  return basicAuth?.split(':');
```
royra, line 17: should make sure there are exactly 2 parts

royra (approved):
LGTM, minor comments

packages/cli/src/commands/up.ts
Yshayy: The problem is that `include-access-credentials` UI is broken, when these flags are on

royra: It's also broken when your branch name is too long. Having proper tables according to the tty column width is a feature - we can decide to never implement it but as long as we do, it should work.

packages/core/src/commands/up/index.ts
Yshayy: It's not in withDockerSocket, it's in the scope of up(...)

royra: Right, sorry

packages/core/src/commands/up/index.ts
Yshayy: Don't think it's needed here. 
In terms of composability, this allows to easily add another manipulation to the URL in addition to addBasicAuthCredentials, so it's more flexible.
In that case, however, short-circuiting with includeAccessCredentials conditional make less sense, but I think it's the right balance.

packages/cli/src/commands/urls.ts
Yshayy: done

packages/cli/src/commands/urls.ts
Yshayy: Refactored a bit, include-access-credentials is no longer passed to core/lib.
The reasoning is that is more about manipulation and formatting to the URLs, I don't think internal "up"/"urls" need to take care of that. (It also removes details, since we no longer have the raw URL of the tunnels).

I think that for plugins, we'll probably want to send both (url, and "secure-url") or alternatively, expose convinence functions for embedded basic-auth credentials in a url

packages/core/src/credentials/index.ts
Yshayy: done

packages/core/src/commands/up/index.ts
Yshayy: I refactored a bit, see previous comment

packages/core/src/commands/urls.ts
Yshayy: Fixed

tunnel-server/src/auth.ts
Yshayy: Looking at the RFC, it seems to allow ":" in the password part. (not the user)
Updated it

Yshayy: Not that it matters, due to our usage of Basic auth header.

PR #110, opened 2023-06-20
kubernetes single pod driver

royra:


packages/driver-kube-docker/src/driver/client/metadata.ts
```diff
@@ -0,0 +1,128 @@
...
+export const ANNOTATIONS = {
+  CREATED_AT: 'preevy.dev/created-at',
+  TEMPLATE_HASH: 'preevy.dev/template-hash',
+  KUBERNETES_KIND: 'preevy.dev/kubernetes-kind',
+  KUERBETES_API_VERSION: 'preev.dev/kubernetes-api-version',
+  DEPLOYMENT_REVISION: 'deployment.kubernetes.io/revision',
+}
+
+export const markObjectAsDeleted = (
+  spec: k8s.KubernetesObject,
```
Yshayy, line 27: spec is a bit confusing due to k8s spec property, maybe resource

packages/driver-kube-docker/src/driver/client/port-forward.ts
```diff
@@ -0,0 +1,57 @@
...
+      .catch(err => {
+        log.debug('forward api error', err)
+        socket.emit('error', err)
+      })
+
+    if (!forwardResult) {
+      return
+    }
+
+    const ws = typeof forwardResult === 'function' ? forwardResult() : forwardResult
```
Yshayy, line 33: what can forwardRequest be here?

packages/driver-kube-docker/src/driver/creation-driver.ts
```diff
@@ -0,0 +1,69 @@
+import { Flags, Interfaces } from '@oclif/core'
+import { MachineCreationDriver, MachineCreationDriverFactory, telemetryEmitter } from '@preevy/core'
+import { DeploymentMachine, machineFromDeployment } from './common'
+import { DriverContext, clientFromConfiguration, machineConnection, flags as machineDriverFlags } from './driver'
+
+type MachineCreationDriverContext = DriverContext & { serverSideApply: boolean }
+
+export const flags = {
```
Yshayy, line 8: Maybe we shouldn't expose it as a flag for now.
This feature might not be widely used or needed, and if we allow it, we'll need to support it.

packages/driver-kube-docker/src/driver/driver.ts
```diff
@@ -0,0 +1,146 @@
...
+      stderr,
+    }
+    return await client.exec(opts)
+  },
+
+  connect: machine => machineConnection(client, machine as DeploymentMachine),
+})
+
+export const flags = {
+  namespace: Flags.string({
```
Yshayy, line 106: Does the namespace need to exists?

packages/driver-kube-docker/src/driver/client/pagination.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const paginationIterator = <
+  T,
+  P extends Page<T> = Page<T>,
+>(fetch: (continueToken?: string) => Promise<P>): AsyncIterableIterator<T> => coreIter(async token => {
+    const r = await fetch(token)
+    return {
+      items: r.body.items,
+      // eslint-disable-next-line no-underscore-dangle
```
Yshayy, line 17: I think it's worth removing this rule:
no-underscore-dangle

site/docs/drivers/kube-docker.md
```diff
@@ -0,0 +1,84 @@
+---
+sidebar_position: 1
+title: Kubernetes Docker-in-Docker Driver
+---
+
+# Kubernetes Docker-in-Docker driver
```
Yshayy, line 6: I think in general, kube-docker might be the wrong name here, as this plugin can use different strategy in the future.
In addition, Docker-In-Docker is not really used. It's just Docker. (affectively, the pod is also not running in Docker)
Like Tekton that uses Docker for building: https://github.com/tektoncd/catalog/tree/main/task/docker-build/0.1

site/docs/roadmap.md
```diff
@@ -9,7 +9,7 @@ sidebar_position: 7
 
 - [x] [AWS lightsail](drivers/aws-lightsail.md)
 - [x] [GCP Compute Engine](https://cloud.google.com/)
-- [ ] [Azure](https://azure.microsoft.com/)
+- [x] [Azure](https://azure.microsoft.com/)
 - [ ] [fly.io](https://fly.io/)
```
Yshayy, line 13: need to add Kuberentes here

Yshayy (commented):
Looks good.
Added some notes.

packages/cli/src/commands/init.ts
```diff
@@ -133,7 +135,27 @@ export default class Init extends BaseCommand {
...
       }
 
-      await this.config.runCommand('profile:create', ['--log-level', this.flags['log-level'] ?? 'info', profileAlias, location, '--driver', driver, ...Object.entries(driverFlags).map(([key, value]) => `--${driver}-${key}=${value}`)])
+      await this.config.runCommand('profile:create', [
+        '--log-level', this.flags['log-level'] ?? 'info',
+        profileAlias,
+        location,
+        '--driver', driver,
+        ...Object.entries(driverFlags)
+          .filter(excludeDefaultFlags(driverStatic.flags as Record<string, Flag<unknown>>))
```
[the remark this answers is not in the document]
royra: yes - otherwise it will be treated as a flag specified by the user

packages/cli/src/commands/init.ts
```diff
@@ -133,7 +135,27 @@ export default class Init extends BaseCommand {
...
+          .map(([key, value]) => {
+            const driverKey = `${driver}-${key}`
+            if (typeof value === 'boolean') {
+              return value ? `--${driverKey}` : `--no-${driverKey}`
+            }
+            if (typeof value === 'string') {
+              return `--${driverKey}=${shellEscape([value])}`
```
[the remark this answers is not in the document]
royra: right, removed

packages/cli/src/commands/shell.ts
```diff
@@ -25,14 +25,18 @@ export default class Shell extends DriverCommand<typeof Shell> {
...
       args: restArgs,
       machineDriver: driver,
       log: this.logger,
     })
 
-    this.exit(code ?? 0)
+    if ('code' in result) {
+      this.exit(result.code)
+    } else {
+      process.kill(process.pid, result.signal)
```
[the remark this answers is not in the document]
royra: if the child process ends with a signal (like ctrl-c), it needs to propagate to the current process.

packages/driver-kube-docker/src/driver/creation-driver.ts
royra: since this feature is new with `kubectl` I feel like we should stay close to its behavior.
same logic can be applied in reverse - we can remove it if it's not used.

packages/driver-kube-docker/src/driver/driver.ts
royra: yes, is it not clear enough?

site/docs/drivers/kube-docker.md
royra: what different strategy do you have in mind?
I wanted to differentiate this driver from another we might have in the future which will use native kubernetes resources.

PR #112, opened 2023-06-26
baseUrl to rootUrl

royra:
replace `baseUrl` object returned from the `hello` command with `rootUrl` string, in a backwards-compatible way

### DIFF
--- a/packages/common/src/ssh/base-client.ts
+++ b/packages/common/src/ssh/base-client.ts
@@ -47,12 +47,6 @@ export type HelloResponse = {
   rootUrl: string
 }
 
-// TODO: temporary, baseUrl should be a replaced with rootUrl after the next deployment of the tunnel service
-type LegacyBaseUrl = { hostname: string; port: string; protocol: string }
-type LegacyHelloResponse = Omit<HelloResponse, 'rootUrl'> & {
-  baseUrl: LegacyBaseUrl
-}
-
 const connectTls = (
   { hostname: host, port, tlsServerName, insecureSkipVerify }: Pick<SshTlsConnectionConfig, 'hostname' | 'port' | 'tlsServerName' | 'insecureSkipVerify'>
 ) => new Promise<TLSSocket>((resolve, reject) => {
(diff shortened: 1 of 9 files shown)


tunnel-server/index.ts
```diff
@@ -69,7 +69,9 @@ const sshServer = createSshServer({
   },
   onHello: (clientId, tunnels) => JSON.stringify({
     clientId,
-    baseUrl: BASE_URL.toString(),
+    // TODO: backwards compat, remove when we drop support for CLI v0.0.34
+    baseUrl: { hostname: BASE_URL.hostname, port: BASE_URL.port, protocol: BASE_URL.protocol },
```
Yshayy, line 73: LGTM :)
Basically 0.0.35 still use the old baseURL I think.

I think we can create the change of the CLI in another PR and push release afterward. (0.0.360 (in the meantime, return the original baseUrl functionility).

There's no need to merge code with branching on baseUrl/RootUrl as the server should support both cases after deployment (and before, it just support v0.0.35)

Yshayy (approved):

packages/common/src/ssh/base-client.ts
```diff
@@ -44,8 +44,13 @@ export const parseSshUrl = (s: string): Pick<SshConnectionConfig, 'hostname' | '
 export type HelloResponse = {
   clientId: string
   tunnels: Record<string, string>
-  // TODO: baseUrl should be a string in the next deployment of the tunnel service, this is for backwards compat
-  baseUrl: string | { hostname: string; port: string; protocol: string }
+  rootUrl: string
+}
+
+// TODO: temporary, baseUrl should be a replaced with rootUrl after the next deployment of the tunnel service
+type LegacyBaseUrl = { hostname: string; port: string; protocol: string }
```
Yshayy, line 51: There's no need to merge code with branching on baseUrl/RootUrl as the server should support both cases after deployment (and before, it just support v0.0.35)

Yshayy (approved):

PR #113, opened 2023-06-29
Small fixes in telemetry and plugin init

### LINKED ISSUE #103: [BUG] - Preevy should print meaningful error if docker executable not found
Currently, preevy print this error:
```
Error: spawn docker ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:476:16)
    at processTicksAndRejections (node:internal/process/task_queues:82:2
```
It's very aggressive and we should print a friendlier message

### LINKED ISSUE #114: [BUG] - Preevy down require docker-compose context to work
Using `preevy down --id="myenv"` should be able to destroy the environment in similar fashion to docker-compose -p

### LINKED ISSUE #115: [BUG] - Preevy should not fail on 'init' when Docker is not installed
Currently, loading plugins is dependent on docker-compose and Preevy crashes when there's no docker client.
Ideally, we should be able to run without plugins most command (maybe add warning), and the behavior should be handled more gracefully.

Yshayy:
- closes #114 
- closes #115
- closes #103 
In addition, this fix a bug of unhandled exception and additional telemetry

### DIFF
--- a/packages/cli/src/commands/down.ts
+++ b/packages/cli/src/commands/down.ts
@@ -28,11 +28,11 @@ export default class Down extends DriverCommand<typeof Down> {
     const log = this.logger
     const { flags } = await this.parse(Down)
     const driver = await this.driver()
-    const userModel = await this.ensureUserModel()
-    const projectName = userModel.name
-    log.debug(`project: ${projectName}`)
-    const envId = flags.id ?? await findAmbientEnvId(projectName)
-
+    const [envId, userModel] = flags.id ? [flags.id, { name: flags.id }] : await (async () => {
+      const model = await this.ensureUserModel()
+      log.debug(`project: ${model.name}`)
+      return [await findAmbientEnvId(model.name), model]
+    })()
     log.debug(`envId: ${envId}`)
     const machine = await driver.getMachine({ envId })
     if (!machine) {
--- a/packages/cli/src/hooks/init/load-plugins.ts
+++ b/packages/cli/src/hooks/init/load-plugins.ts
@@ -5,7 +5,7 @@ import { telemetryEmitter } from '@preevy/core'
 const wrappedHook: OclifHook<'init'> = async function wrappedHook(...args) {
   try {
     const { id } = args[0]
-    if (id === 'help' || id === 'version' || id === 'init') {
+    if (id === 'version' || id === 'init') {
       return
     }
     await initHook.call(this, ...args)
--- a/packages/cli/src/hooks/init/load-plugins.ts
+++ b/packages/cli/src/hooks/init/load-plugins.ts
@@ -1,13 +1,18 @@
 import { Hook as OclifHook } from '@oclif/core'
 import { initHook } from '@preevy/cli-common'
+import { telemetryEmitter } from '@preevy/core'
 
 const wrappedHook: OclifHook<'init'> = async function wrappedHook(...args) {
   try {
+    const { id } = args[0]
+    if (id === 'help' || id === 'version' || id === 'init') {
+      return
+    }
     await initHook.call(this, ...args)
   } catch (e) {
     // eslint-disable-next-line no-console
-    console.error('init hook error', e)
-    process.exit(1)
+    console.warn(`warning: failed to init plugins: ${e}`)
+    telemetryEmitter().capture('plugin-init-error', { error: `${e}` })
   }
 }
 
--- a/packages/cli/src/hooks/init/load-plugins.ts
+++ b/packages/cli/src/hooks/init/load-plugins.ts
@@ -11,8 +11,10 @@ const wrappedHook: OclifHook<'init'> = async function wrappedHook(...args) {
     await initHook.call(this, ...args)
   } catch (e) {
     // eslint-disable-next-line no-console
-    console.warn(`warning: failed to init plugins: ${e}`)
+    console.warn(`warning: failed to init context: ${e}`)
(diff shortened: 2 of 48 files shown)


packages/cli/src/commands/down.ts
```diff
@@ -28,11 +28,11 @@ export default class Down extends DriverCommand<typeof Down> {
     const log = this.logger
     const { flags } = await this.parse(Down)
     const driver = await this.driver()
-    const userModel = await this.ensureUserModel()
-    const projectName = userModel.name
-    log.debug(`project: ${projectName}`)
-    const envId = flags.id ?? await findAmbientEnvId(projectName)
-
+    const [envId, userModel] = flags.id ? [flags.id, { name: flags.id }] : await (async () => {
```
royra, line 31: `envId` is incorrect here - missing branch suffix.
what is this fixing?

packages/cli/src/hooks/init/load-plugins.ts
```diff
@@ -1,12 +1,19 @@
 import { Hook as OclifHook } from '@oclif/core'
 import { initHook } from '@preevy/cli-common'
+import { telemetryEmitter } from '@preevy/core'
 
 const wrappedHook: OclifHook<'init'> = async function wrappedHook(...args) {
   try {
+    const { id } = args[0]
```
royra, line 7: IMO this function shouldn't include any logic besides wrapping with try..catch, you can move the id check to the hook itself.

packages/core/src/child-process.ts
```diff
@@ -71,7 +72,7 @@ export const childProcessStdoutPromise = async (
   return orderedOutput(output).stdout().toString('utf-8').trim()
 }
 
-export const spawnPromise = (
+export const spawnPromise = async (
```
royra, line 75: redundant `async`

packages/core/src/compose/client.ts
```diff
@@ -35,21 +41,26 @@ const composeClient = (
   executer: Executer,
   composeFiles: string[] | Buffer,
 ) => {
-  const execComposeCommand = (args: string[]) => executer({
+  const execComposeCommand = async (args: string[]) => executer({
```
royra, line 44: redundant `async`

packages/core/src/compose/client.ts
```diff
@@ -116,11 +127,11 @@ export const localComposeClient = (
     ...args: Parameters<typeof spawnComposeArgs>
   ) => addStdIn(spawn(...spawnComposeArgs(...args)))
 
-  const spawnComposePromise = (
+  const spawnComposePromise = async (
```
royra, line 130: redundant `async`

packages/core/src/compose/client.ts
```diff
@@ -116,11 +127,11 @@ export const localComposeClient = (
...
     ...args: Parameters<typeof spawnComposeArgs>
   ) => childProcessPromise(spawnCompose(...args))
 
-  const executer: Executer = ({ args }) => childProcessStdoutPromise(spawnCompose(args, {}))
+  const executer: Executer = async ({ args }) => childProcessStdoutPromise(spawnCompose(args, {}))
```
royra, line 134: redundant `async`

packages/core/src/telemetry/machine-id.ts
```diff
@@ -33,12 +34,13 @@ const calcMachineId = async () => {
 }
 
 export const memoizedMachineId = async (dataDir: string) => {
+  const dir = localFs(dataDir)
```
royra, line 37: why is `localFs` needed here?

royra (commented):
LGTM, see minor comments

royra (approved):

packages/cli/src/commands/down.ts
Yshayy: The case of using down --id without docker/compose. (#114)

royra: if the project name is only needed for the `debug.log` call, I suggest removing it to simplify:
```suggestion
    const envId = flags.id ?? await findAmbientEnvId((await this.ensureUserModel()).name)
```

packages/core/src/child-process.ts
Yshayy: The async isn't redundant, if `spawn` logic throw the error is not wrapped.
Ideally, any function that return a Promise should give a strong guarantee that the exception is always wrapped in the promise return value. 
`async` force that, for example:
```
await ((function(){ throw "abc"; return Promise.resolve(5)}()).catch(e=> { }))
```
throws "abc"

```
await ((async function(){ throw "abc"; return Promise.resolve(5)}()).catch(e=> { }))
```
doesn't throw

packages/core/src/telemetry/machine-id.ts
Yshayy: in container environments, datadir folder didn't exist

packages/cli/src/commands/down.ts
Yshayy: That was the variable before.
```const envId = flags.id ?? await findAmbientEnvId(projectName)```

Yshayy:
Went with `standard` approach:
https://github.com/standard/standard/issues/1442
https://github.com/standard/eslint-config-standard-with-typescript/pull/253

Remote no-return-await (to avoid synchronous throws and use async stack-traces: https://v8.dev/blog/fast-async#improved-developer-experience) and force typescript to require the use of `await` when using async

Issue #120, opened 2023-07-05
Add context option to kubernetes provider


Yshayy:
To ensure we're using the right cluster, we should initialize the profile with kube-context (while still allowing users to override it in up/down command)

PR #125, opened 2023-07-09
Update kube-pod driver

Yshayy:
Add support for context.
Update Init questions/answers.
Solve #120 

Should be merged after #126

### DIFF
--- a/packages/driver-kube-pod/src/driver/client/index.ts
+++ b/packages/driver-kube-pod/src/driver/client/index.ts
@@ -57,15 +57,19 @@ const ensureSingleDockerHostDeployment = (): ApplyFilter => {
   }
 }
 
-const kubeClient = ({ log, namespace, kc, profileId, template, package: packageDetails, kubeconfig }: {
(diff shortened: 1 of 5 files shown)


packages/driver-kube-pod/src/driver/client/index.ts
```diff
@@ -57,15 +57,19 @@ const ensureSingleDockerHostDeployment = (): ApplyFilter => {
...
   log: Logger
   kc: k8s.KubeConfig
   kubeconfig?: string
+  context?: string
   namespace: string
   profileId: string
   template: Buffer | string | Promise<Buffer | string>
   package: Package | Promise<Package>
 }) => {
+  if (context) {
```
royra, line 70: unexpected side effect, better to set in caller

royra (approved):

packages/driver-kube-pod/src/driver/client/index.ts
Yshayy: done

PR #126, opened 2023-07-09
Add support for configuration of profile

### LINKED ISSUE #23: Update profile UX to enable viewing and setting default value after profile init
Currently, only the init or create command allow to set default flags (such as region) and they are only used on initialization of a profile.

Yshayy:
Fix #23 

Add the command `preevy profile configure`:

Usage examples:
```
# View current configuration
preevy profile config view ->

Current configuration for kube-pod:  
namespace:     test123
server-side-apply: false


# Update namespace for kube-pod driver
preevy profile config update --kube-pod-namespace=abc ->

Updated profile configuration
Current configuration for kube-pod:
namespace:     abc
server-side-apply: false


# Update namespace for kube-pod driver
preevy profile config update --unset=kube-pod-server-side-apply ->
# Or
preevy profile config update --unset=server-side-apply ->

Updated profile configuration
Current configuration for kube-pod:
namespace:     abc
```

### DIFF
--- a/packages/cli/src/commands/profile/config/update.ts
+++ b/packages/cli/src/commands/profile/config/update.ts
@@ -9,8 +9,8 @@ import {
   machineCreationflagsForAllDrivers,
   DriverName,
   machineDrivers
-} from '../../drivers'
-import ProfileCommand from '../../profile-command'
+} from '../../../drivers'
+import ProfileCommand from '../../../profile-command'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const removeDefaultFlagsDef = <T extends {[key: string]: Flag<any> } >(flags: T): T =>
@@ -21,7 +21,7 @@ const removeDefaultFlagsDef = <T extends {[key: string]: Flag<any> } >(flags: T)
   }) as T
 
 // eslint-disable-next-line no-use-before-define
-export default class ConfigureProfile extends ProfileCommand<typeof ConfigureProfile> {
+export default class UpdateProfileConfig extends ProfileCommand<typeof UpdateProfileConfig> {
   static description = 'View and update profile configuration'
 
   static flags = {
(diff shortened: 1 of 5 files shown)


royra (commented):
A few ideas about the UX.

- It's weird that a command named "configure" also does listing.
- Profile has more in it than flags
- There needs to be a way to redo interactive profile init without recreating the public key

How about:

```bash
# list flags
preevy profile flags ls [--profile=profile-alias]

# set flag(s)
preevy profile flags set [--profile=profile-alias] key1=value key2=value

# unset flag(s)
preevy profile flags unset [--profile=profile-alias] key1 key2

# just like `init`, with defaults set to the current values
preevy profile configure [--profile=profile-alias]
```

royra:
> Good idea, I think in general, we should always allow specifying the `--profile` flag.

Do you mean making it required? IMO the current profile is a good default.

> In terms of the `flags` command name, it's a bit too technical, I think.

Maybe "defaults" then?

> The problem with `set key1=key2` is that it's not part of the flags and doesn't appear in the help command like the flags.

We can make the `args` help return the driver flags' help. Help can be [dynamic](https://github.com/oclif/core/blob/main/src/help/index.ts#L108) and depend on the state stored in the profile (i.e, the `driver` flag) to show only the chosed driver's flags.

> What do you think about: `preevy profile config` -> show profile config.  `preevy profile config update {flags} {or --set}` -> update and print new config. `preevy profile config reset` -> reset default values and prompt the interactive init.

- I'm not sure what "profile config" means, it's too generic and it covers stuff that is not default flags. See suggestion about "defaults".
- It's confusing that there is an implicit "list" subcommand here. When a subcommand is not specified it lists the defaults, but there is also an explicit "update" subcommand?
- Reset and start from scratch is a bad UX for changing a single option. As a user I would rather mash the enter key on all questions to leave my answers as is as is until I get to the one thing I want to change.

Issue #123, opened 2023-07-05
Support authentication session in tunnel server


Yshayy:
We should be able to support the session for:
- Support flows different from basic authentication
- Improving the experience of the login between different environments/services (no need to re-login)
- Allow to create a session without passing the authorization header the can conflict with internal app authorization header and doesn't work properly when using websockets

PR #127, opened 2023-07-09
Add initial session support in tunnel server

Yshayy:
Implement #123:
- [x] Add support for cookie mechanism
- [x] Allow to extract JWT from different mechanisms (Bearer, Basic, Cookie)
- [x] cookie signing
- [ ] better authentication redirect flow (pass basicAuth param)


tunnel-server/src/app.ts
```diff
@@ -31,4 +40,37 @@ export const app = ({ isProxyRequest, proxyHandlers, logger }: {
...
+            returnPath: { type: 'string' }
+          },
+          required: ['env']
+        }
+      }
+    }, async function (req, res) {
+      const {env:envId, returnPath = '/'} = req.query
+      if (!returnPath.startsWith('/')){
+        res.statusCode = 400
+        return {error: 'returnPath must be a relative path' }
```
royra, line 57: nit: relative path does not start with `/`
```suggestion
        return {error: 'returnPath must be an absolute path' }
```

tunnel-server/src/app.ts
```diff
@@ -31,4 +40,37 @@ export const app = ({ isProxyRequest, proxyHandlers, logger }: {
...
+      if (!session.user){
+        const auth = authenticator([JwtAuthenticator(env)])
+        const result = await auth(req.raw)
+        if (!result.isAuthenticated){
+          return unauthorized(res.raw)
+        }
+        session.set(result.claims)
+        session.save()
+      }
+      return res.redirect(new URL(returnPath, replaceHostname(baseUrl, `${envId}.${baseUrl.hostname}`)).toString())
```
royra, line 74: missing `clientId`?

tunnel-server/src/proxy.ts
```diff
@@ -54,17 +69,33 @@ export function proxyHandlers({
...
+            res.end();
+            return;
           }
-        } catch (error) {
-          res.statusCode = 400;
-          res.end();
-          return;
+        }
+
+        if (session.user?.role !== 'admin') {
```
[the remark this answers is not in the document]
Yshayy: It's more for using the SaaS, I believe you can be authenticated (for example, different org), but without permissions.
There's currently only one role 'admin' and one scope 'admin' and the authentication cookie session is scoped to a profile. 
Yet, I still think it's better to keep the interface more open, the cost is relatively low, and it makes it more forward compatible.

tunnel-server/src/app.ts
Yshayy: I think the envId contains the clientId (since it's basically the host we're comparing).
We need to change the name of envStore, it's not really env it's a hostname->socket/tunnel

tunnel-server/src/app.ts
royra: understood - still not relative, error is misleading
```suggestion
        return {error: 'returnPath must not be a full URL' }
```

tunnel-server/src/app.ts
```diff
@@ -3,20 +3,29 @@ import { fastifyRequestContext } from '@fastify/request-context'
...
 }) =>
   Fastify({
     serverFactory: (handler) => {
       const {wsHandler:proxyWsHandler, handler: proxyHandler } = proxyHandlers
       const server = http.createServer((req, res) => {
-        if (isProxyRequest(req)){
-          return proxyHandler(req, res)
-        }
-        return handler(req, res)
+        return match(req)
```
royra, line 25: Not sure using a custom DSL like `match` is better than plain `if`s, matter of style I guess.

Also - the first `match` call can be omitted (since it's covered by the `otherwise` call)

tunnel-server/src/auth.ts
```diff
@@ -1,55 +1,123 @@
...
+  password: string
+}
+
+export type BearerAuthorizationHeader = {
+  scheme: 'Bearer'
+  token: string
 }
 
-function extractBasicAuth(req: IncomingMessage){
+const cookieName = 'preevy-sass-jwt'
```
royra, line 44: typo
```suggestion
const cookieName = 'preevy-saas-jwt'
```

tunnel-server/src/auth.ts
```diff
@@ -1,55 +1,123 @@
...
+        const thumbprint = await calculateJwkThumbprintUri(await exportJWK(env.publicKey))
+        return {
+          publicKey: env.publicKey,
+          issuer: `preevy://${thumbprint}`,
+          extractClaims: (token:JWTPayload)=>({
+            role: 'admin',
+            type: 'profile',
+            exp: token.exp,
+            scopes: ['admin'],
+            sub: `preevy-profile:${env.publicKeyThumbprint}`
```
royra, line 87: this is not exactly the profile but rather the public key thumbprint
```suggestion
            sub: `preevy-pubkey:${env.publicKeyThumbprint}`
```

tunnel-server/src/proxy.ts
```diff
@@ -54,17 +69,33 @@ export function proxyHandlers({
...
+            if (!authResult.isAuthenticated){
+              return unauthorized(res);
+            }
+            session.set(authResult.claims)
+            if (authResult.login && req.method === 'GET'){
+              session.save()
+              redirectToLogin(res, env.hostname, req.url)
+              return;
+            } 
+            if (authResult.method.type === 'header'){
```
royra, line 87: this is a smelly side-effect, how about having the authenticator transform the headers?

tunnel-server/src/proxy.ts
```diff
@@ -96,9 +127,11 @@ export function proxyHandlers({
       }
 
       if(env.access === 'private'){
-        // need to support session cookie, native browser Websocket api doesn't forward authorization header
-        socket.end();
-        return;
+        const session = sessionManager(req, undefined as any, env.clientId)
+        if (session.user?.role !== 'admin'){
+          socket.end();
```
royra, line 132:
```suggestion
          socket.end('Invalid role');
```

tunnel-server/src/seesion.ts
```diff
@@ -0,0 +1,35 @@
...
+import * as z from "zod"
+
+// for testing, for production workload use the env var COOKIE_SECRETS
+function generateSecret(){
+  return randomBytes(32)
+    .toString('base64')
+    .slice(0, 32)
+}
+
+export function sessionManager<T>(opts: {domain: string, schema: z.ZodSchema<T>, keys?: string[] }){
```
royra, line 13: nit: `Xmanager` is a naming anti-pattern, how about just `session`?

royra (commented):
some more comments, needs rebase, otherwise LGTM

royra (approved):

tunnel-server/src/seesion.ts
Yshayy: It's to represent a mechanism for creating a factory for extracting a stateful session from a request, so simply 'session' is less suited here.
Manager makes sense IMO, but I'm open to a less generic name.

tunnel-server/src/proxy.ts
Yshayy: It can be 'forbidden' or 'unauthorized', but I'm not sure how the client can consume it.

PR #128, opened 2023-07-10
clean up tunnel server SOC

royra:
can review the two commits separately

- refactor tunnel server
- converted from callbacks to events


tunnel-server/src/ssh-server.ts
```diff
@@ -4,196 +4,265 @@ import net from 'net'
...
+                      socket.end()
+                      socketServer.close((err) => {
+                        log.error('error closing socket server for request %j: %j', request, inspect(err))
+                      })
+                      return
+                    }
+                    upstream.pipe(socket).pipe(upstream)
+                  }
+                )
+              })
```
Yshayy, line 197: Reject is missing here, but expected in the signature

tunnel-server/src/ssh-server.ts
```diff
@@ -4,196 +4,265 @@ import net from 'net'
...
+            preevySshClient.emit('hello', {
+              stdout: channel.stdout,
+              exit: (status: number) => {
+                channel.stdout.exit(status)
+                if (tunnels.size === 0) {
+                  channel.close()
+                }
+              },
+            })
+            // channel.stdout.on('close', () => {
```
Yshayy, line 253: Comment can be removed?

tunnel-server/src/ssh-server.ts
```diff
@@ -4,196 +4,265 @@ import net from 'net'
...
-            channel.close()
-          }
         })
-      })
-  }
-)
+    }
+  )
+
+  return Object.assign(serverEmitter, {
```
Yshayy, line 264: nit - might be clearer to return the server and assign the "on:" server Emitter.on
Since the return value is basically SSH server with an event handler.

tunnel-server/src/ssh-server.ts
Yshayy: Probabbly the behavior should be calling the reject as well

Issue #144, opened 2023-07-26
[Improvement]: Add messaging for initial CLI cloud interactions


Yshayy:
### What feature?

When using `preevy up` or other commands, it can take several seconds until we see the output due to connectivity with cloud providers. (for example, downloading profile)
We can improve this experience by adding console messages that log these interactions.

### Record

- [X] I agree to follow this project's Code of Conduct

Yshayy:
Yes, we'll be happy to accept contribution for it.
@roy9495 let me know if you need more details, I'm also available on our Slack channel:
https://join.slack.com/t/preevyusers/shared_invite/zt-1shi7t2dy-~glGUC5pWCtGrH19bKNXrQ

PR #148, opened 2023-07-28
Messaging for initial CLI cloud interaction added

roy9495:
Fix: #144 

Console messages have been added for the "preevy up" command to enhance the user experience.

### DIFF
--- a/packages/cli/src/commands/up.ts
+++ b/packages/cli/src/commands/up.ts
@@ -5,6 +5,7 @@ import {
 } from '@preevy/core'
 import { tunnelServerFlags } from '@preevy/cli-common'
 import { inspect } from 'util'
+import { withSpinner } from '@preevy/core/src/spinner'
 import { tunnelServerHello } from '../tunnel-server-client'
 import MachineCreationDriverCommand from '../machine-creation-driver-command'
 import { envIdFlags, urlFlags } from '../common-flags'
@@ -51,11 +52,14 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
       insecureSkipVerify: flags['insecure-skip-verify'],
     }
 
-    const { hostKey, clientId, rootUrl } = await tunnelServerHello({
-      tunnelingKey,
-      knownServerPublicKeys: pStore.knownServerPublicKeys,
-      tunnelOpts,
-      log: this.logger,
+    const { clientId, rootUrl, hostKey } = await withSpinner(async spinner => {
+      spinner.text = 'Getting tunnel server details...'
+      return await tunnelServerHello({
+        tunnelingKey,
+        knownServerPublicKeys: pStore.knownServerPublicKeys,
+        tunnelOpts,
+        log: this.logger,
+      })
     })
 
     telemetryEmitter().group({ type: 'profile' }, { proxy_client_id: clientId })
@@ -85,19 +89,22 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
 
     this.log(`Preview environment ${envId} provisioned at: ${machine.locationDescription}`)
 
-    const flatTunnels = await commands.urls({
-      rootUrl,
-      clientId,
-      envId,
-      tunnelingKey,
-      includeAccessCredentials: flags['include-access-credentials'],
-      showPreevyService: flags['show-preevy-service-urls'],
-      retryOpts: {
-        minTimeout: 1000,
-        maxTimeout: 2000,
-        retries: 10,
-        onFailedAttempt: e => { this.logger.debug(`Failed to query tunnels: ${inspect(e)}`) },
-      },
+    const flatTunnels = await withSpinner(async spinner => {
+      spinner.text = 'Getting tunnel URLs...'
+      return await commands.urls({
+        rootUrl,
+        clientId,
+        envId,
+        tunnelingKey,
+        includeAccessCredentials: flags['include-access-credentials'],
+        showPreevyService: flags['show-preevy-service-urls'],
+        retryOpts: {
+          minTimeout: 1000,
+          maxTimeout: 2000,
+          retries: 10,
+          onFailedAttempt: e => { this.logger.debug(`Failed to query tunnels: ${inspect(e)}`) },
+        },
+      })
     })
 
     const urls = await filterUrls({


royra (commented):
Thanks for the PR! See comments.

roy9495:
I tried to do the things that you asked for. Please review the commit.

packages/cli/src/commands/up.ts
```diff
@@ -51,73 +52,82 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
       insecureSkipVerify: flags['insecure-skip-verify'],
     }
 
-    const { hostKey, clientId, rootUrl } = await tunnelServerHello({
-      tunnelingKey,
-      knownServerPublicKeys: pStore.knownServerPublicKeys,
-      tunnelOpts,
-      log: this.logger,
+    await withSpinner(async spinner => {
```
royra, line 55: spinner and log don't work together well, because the logs print a newline which is left behind on the next "frame" of the spinner.

also - the spinner should only wrap the `Initializing tunnel connection...` action, not the whole flow.

packages/cli/src/commands/up.ts
```diff
@@ -51,14 +52,26 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
     }
 
-    const { hostKey, clientId, rootUrl } = await tunnelServerHello({
-      tunnelingKey,
-      knownServerPublicKeys: pStore.knownServerPublicKeys,
-      tunnelOpts,
-      log: this.logger,
+    let clientId = ''
+    let rootUrl = ''
+    let hostKey: Buffer = Buffer.from('')
```
royra, line 57: no need to use `let` here. `withSpinner` returns whatever its given function returns, so you could return the result of `tunnelServerHello`.

packages/cli/src/commands/up.ts
```diff
@@ -51,14 +52,26 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
+
+    await withSpinner(async spinner => {
+      spinner.start('Initializing tunnel connection...')
```
royra, line 60: there's no need to call `spinner.start`, `withSpinner` accepts a `text` option (part of the Ora options)

packages/cli/src/commands/up.ts
```diff
@@ -51,14 +52,26 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
+        tunnelingKey,
+        knownServerPublicKeys: pStore.knownServerPublicKeys,
+        tunnelOpts,
+        log: this.logger,
+      })
+      clientId = result.clientId
+      rootUrl = result.rootUrl
+      hostKey = result.hostKey
+      telemetryEmitter().group({ type: 'profile' }, { proxy_client_id: clientId })
+      spinner.succeed('Tunnel connection established.')
```
royra, line 71: no need to call `spinner.succeed`, `withSpinner` accepts `succesText`

packages/cli/src/commands/up.ts
```diff
@@ -51,14 +52,26 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
+      const result = await tunnelServerHello({
+        tunnelingKey,
+        knownServerPublicKeys: pStore.knownServerPublicKeys,
+        tunnelOpts,
+        log: this.logger,
+      })
+      clientId = result.clientId
+      rootUrl = result.rootUrl
+      hostKey = result.hostKey
+      telemetryEmitter().group({ type: 'profile' }, { proxy_client_id: clientId })
```
royra, line 70: why is this line part of the spinner callback?

packages/cli/src/commands/up.ts
```diff
@@ -99,6 +113,8 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
       },
     })
 
+    this.log('Tunnels queried successfully.')
```
royra, line 116: this is redundant with `successText`

packages/cli/src/commands/up.ts
royra: the text here is misleading - there is no tunnel "connection".

royra (changes requested):
Most of the comments relate to existing patterns in the code base in not being followed.

roy9495:
Hello sir, I tried to do it like this but clientId, rootUrl, hostKey can't be accessed by other functions outside of withSpinner and the successText line is also giving an error ------------------
~~~
await withSpinner(async spinner => {
      spinner.text = 'Initializing tunnel connection...'
      const { clientId, rootUrl, hostKey } = await tunnelServerHello({
        tunnelingKey,
        knownServerPublicKeys: pStore.knownServerPublicKeys,
        tunnelOpts,
        log: this.logger,
      })
      successText : 'Tunnel established.'
      return { clientId, rootUrl, hostKey }
    })
~~~

packages/cli/src/commands/up.ts
```diff
@@ -51,15 +52,20 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
     }
 
-    const { hostKey, clientId, rootUrl } = await tunnelServerHello({
-      tunnelingKey,
-      knownServerPublicKeys: pStore.knownServerPublicKeys,
-      tunnelOpts,
-      log: this.logger,
-    })
+    const { clientId, rootUrl, hostKey } = await withSpinner(async spinner => {
+      spinner.text = 'Initializing tunnel connection...'
```
royra, line 56: As [mentioned](https://github.com/livecycle/preevy/pull/148#discussion_r1282665645) this is misleading, there is no "connection", it's just a single API call.

```suggestion
      spinner.text = 'Getting tunnel server details'
```

packages/cli/src/commands/up.ts
```diff
@@ -51,15 +52,20 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
+      return await tunnelServerHello({
+        tunnelingKey,
+        knownServerPublicKeys: pStore.knownServerPublicKeys,
+        tunnelOpts,
+        log: this.logger,
+      })
+    }, { successText: 'Tunnel established.' })
```
royra, line 63: there is no tunnel being established.

`successText` is optional, can be removed here and it will show a `✔` next to the previous message.

```suggestion
    })
```

packages/cli/src/commands/up.ts
```diff
@@ -51,15 +52,20 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
...
 
     telemetryEmitter().group({ type: 'profile' }, { proxy_client_id: clientId })
 
+    this.log('Provisioning preview environment...')
```
royra, line 67: This message does not add any information, it should be removed.

```suggestion
```

packages/cli/src/commands/up.ts
```diff
@@ -84,6 +90,7 @@ export default class Up extends MachineCreationDriverCommand<typeof Up> {
     })
 
     this.log(`Preview environment ${envId} provisioned at: ${machine.locationDescription}`)
+    this.log('Querying tunnels for the preview environment...')
```
royra, line 93:
```suggestion
    this.log('Getting tunnel URLs...')
```

royra, line 93: This can be wrapped in a spinner as well.

royra (changes requested):

roy9495:
Thank you for being so patient. I tried to do what you asked for.

royra (approved):

roy9495:
Thank you

PR #150, opened 2023-07-31
Create binaries from Preevy and GH release

Yshayy:
This PR uses Vercel/pkg from bundling binaries.
Due to performance issue in the CI when running arm64 workloads with QEMU, I set up remote docker execution context using `depot`.
It's worth noticing that in any case, we're building the application in Linux environment, so it might have issues with node native modules/add-ons.
The alternative can be using windows/mac/linux builders in actions, but there are limitations there (no macOS arm64).
Currently, I built it only on Linux/macOS platforms because it's easy to verify correctness

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -1,18 +1,35 @@
-FROM node:18-bullseye as builder
-
+FROM node:18-bullseye as base
+FROM base as deps-files
 WORKDIR /app
-COPY --link . /app/
+COPY --link yarn.lock package.json .
+COPY --link packages packages
+RUN find . -type f -not -iname "yarn.lock" -not -iname "package.json" -delete && find . -type l -delete && find . -type d -empty -delete 
+
+FROM alpine as ldid
+RUN wget -q https://github.com/ProcursusTeam/ldid/releases/download/v2.1.5-procursus7/ldid_linux_aarch64 -O /usr/bin/ldid
+RUN chmod +x /usr/bin/ldid
+
+FROM base as builder
+WORKDIR /app
+COPY --link --from=deps-files /app /app
 RUN --mount=type=cache,id=livecycle/preevy-cli/yarn-cache,target=/yarn/cache \
   yarn --cache-folder=/yarn/cache
-
-RUN yarn clean && yarn build
+COPY --link . .
+RUN yarn build
 
 FROM builder as pkg
+ARG CLI_TARGET=macos-arm64
 WORKDIR /app/packages/cli
 ENV PKG_CACHE_PATH=/pkg/cache
+COPY --from=ldid --link /usr/bin/ldid /usr/bin/ldid
 RUN --mount=type=cache,id=livecycle/preevy-cli/pkg-cache,target=/pkg/cache \
-  yarn pkg --public --public-packages tslib --options max_old_space_size=4096 -t node18-alpine-arm64 .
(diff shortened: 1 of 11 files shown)


Dockerfile.cli
```diff
@@ -0,0 +1,35 @@
+FROM node:18-bullseye as base
+FROM base as deps-files
+WORKDIR /app
+COPY --link yarn.lock package.json .
+COPY --link packages packages
+RUN find . -type f -not -iname "yarn.lock" -not -iname "package.json" -delete && find . -type l -delete && find . -type d -empty -delete
```
royra, line 6: instead of copying the entire source tree and then deleting everything except package.json

```suggestion
COPY --link packages/*/package.json packages
```

Dockerfile.cli
```diff
@@ -0,0 +1,35 @@
+FROM node:18-bullseye as base
+FROM base as deps-files
+WORKDIR /app
+COPY --link yarn.lock package.json .
+COPY --link packages packages
+RUN find . -type f -not -iname "yarn.lock" -not -iname "package.json" -delete && find . -type l -delete && find . -type d -empty -delete 
+
+FROM alpine as ldid
```
royra, line 8:
```suggestion
FROM busybox as ldid
```

packages/cli/package.json
```diff
@@ -82,6 +83,10 @@
     },
     "topicSeparator": " "
   },
+  "pkg": {
+    "assets": ["../*/static/**/*", "../compose-tunnel-agent/out/**/*", "../compose-tunnel-agent/package.json", "../compose-tunnel-agent/Dockerfile", "../compose-tunnel-agent/.dockerignore" ],
```
royra, line 87: nit: can we have newlines here? :)

packages/core/src/compose-tunnel-agent-client.ts
```diff
@@ -10,30 +12,52 @@ import { withBasicAuthCredentials } from './url'
...
 const COMPOSE_TUNNEL_AGENT_DIR = path.join(path.dirname(require.resolve('@preevy/compose-tunnel-agent')), '..')
-
-const baseDockerProxyService: ComposeService = {
-  build: {
-    context: COMPOSE_TUNNEL_AGENT_DIR,
-  },
-  labels: {
-    'preevy.access': 'private',
-  },
+declare let process : {
```
royra, line 15: extract to separate module

packages/core/src/compose-tunnel-agent-client.ts
```diff
@@ -10,30 +12,52 @@ import { withBasicAuthCredentials } from './url'
...
 }
 
-export const minimalModelWithDockerProxyService = (name: string): ComposeModel => ({
-  name,
-  services: {
-    [COMPOSE_TUNNEL_AGENT_SERVICE_NAME]: baseDockerProxyService,
-  },
+const pkgComposeTunnelAgentDirFromSnapshot = memoize(() => {
+  // can't use fs.cpSync because it's not patched by pkg (https://github.com/vercel/pkg/blob/bb042694e4289a1cbc530d2938babe35ccc84a93/prelude/bootstrap.js#L600)
+  const copyDirRecursive = (sourceDir: string, targetDir:string) => {
```
royra, line 21: extract to same separate module

royra, line 21:
```suggestion
  const copyDirRecursiveSync = (sourceDir: string, targetDir:string) => {
```

royra (approved):
Nice! See minor comments

Issue #102, opened 2023-06-05
Implement store based on Azure blob storage for profile store


Yshayy:
Implement a profile storage for azure blob storage.

Should be similar to other implementations:
- GS - https://github.com/livecycle/preevy/blob/main/packages/driver-gce/src/fs/index.ts
- S3 - https://github.com/livecycle/preevy/blob/main/packages/driver-lightsail/src/fs/index.ts

Registration is done here:
https://github.com/livecycle/preevy/blob/main/packages/cli/src/fs.ts

Url format should follow conventions of golang cloud sdk:
https://gocloud.dev/howto/blob/#azure

hassansuhaib:
I want to work on this

royra:
Done long ago #423

PR #156, opened 2023-08-02
Feature/azure blob storage for profile store

hassansuhaib:
Issue: https://github.com/livecycle/preevy/issues/102

Implemented Azure Blob storage for Profile Store

This approach requires these environment variables to be set by the user: `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET` because we can't use other methods for creating a blob service account as they require different keys and we only get URL as a param.

### DIFF
--- a/packages/cli/src/fs.ts
+++ b/packages/cli/src/fs.ts
@@ -6,6 +6,9 @@ import inquirerAutoComplete from 'inquirer-autocomplete-standalone'
 import { DriverName } from './drivers.js'
 import ambientAwsAccountId = awsUtils.ambientAccountId
 
+inquirer.registerPrompt('autocomplete', inquirerAutoComplete)
+import { azureStorageBlobFs } from '@preevy/driver-azure'
+
 export const fsFromUrl = async (url: string, localBaseDir: string) => {
   const fsType = fsTypeFromUrl(url)
   if (fsType === 'local') {
@@ -20,6 +23,11 @@ export const fsFromUrl = async (url: string, localBaseDir: string) => {
     // eslint false positive here on case-sensitive filesystems due to unknown type
     // eslint-disable-next-line @typescript-eslint/return-await
     return await googleCloudStorageFs(url)
+  }
+  if (fsType === 'az') {
+    // eslint false positive here on case-sensitive filesystems due to unknown type
+    // eslint-disable-next-line @typescript-eslint/return-await
+    return await azureStorageBlobFs(url)
   }
   throw new Error(`Unsupported URL type: ${fsType}`)
 }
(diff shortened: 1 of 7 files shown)


packages/driver-azure/src/fs/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+    delete: async (filename: string) => {
+      const blockBlobClient = containerClient.getBlockBlobClient(filename)
+      const options: BlobDeleteOptions = {
+        deleteSnapshots: 'include',
+      }
+      const blobDeleteIfExistsResponse: BlobDeleteIfExistsResponse = await blockBlobClient.deleteIfExists(options)
+
+      if (blobDeleteIfExistsResponse.errorCode) {
+        throw new Error(`Error: ${blobDeleteIfExistsResponse.errorCode}`)
```
royra, line 85: according to [this](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/storage/storage-blob/test/blobclient.spec.ts#L344), the `errorCode` will be set if the blob is not found - the contract requires to not throw.

packages/driver-azure/src/fs/index.ts
```diff
@@ -0,0 +1,89 @@
...
+      const blobClient = containerClient.getBlobClient(filename)
+      const result = await blobClient.download()
+      if (result.readableStreamBody !== undefined) {
+        return await streamToBuffer(result.readableStreamBody)
+      }
+      return undefined
+    },
+    write: async (filename: string, content: Buffer | string) => {
+      const blockBlobClient = containerClient.getBlockBlobClient(filename)
+      await blockBlobClient.upload(content, content.length)
```
royra, line 75: what happens if the uploaded blob already exists? should overwrite.

royra (changes requested):
Thanks for your contribution!

- Some comments below to address
- Make sure the code is tested manually, or add automated tests
- Add the option to store the profile on Azure in the [`init`](https://github.com/livecycle/preevy/blob/42711d715/packages/cli/src/commands/init.ts#L91) command
- Add to the docs

packages/cli/src/fs.ts
```diff
@@ -17,5 +18,10 @@ export const fsFromUrl = async (url: string, localBaseDir: string) => {
     // eslint-disable-next-line @typescript-eslint/return-await
     return await googleCloudStorageFs(url)
   }
+  if (fsType === 'az') {
```
Yshayy, line 21: Thanks for the contribution!

The type of the fsType as well as the URL should use `azblob` protocol and be conforming to this format:
https://gocloud.dev/howto/blob/#azure

#102

hassansuhaib: I saw this link in the issue but I got confused as all of the Azure documentation was using `https://`. I guess I'll have to understand the whole architecture of Preevy deeply in order to figure this out.

royra:
fixed by #423

PR #163, opened 2023-08-06
fix ls showing other resources

### LINKED ISSUE #159: Wrong driverId name and no envId on preevy ls
After I tried to use preevy up command to deploy my current changes, preevy created machines on AWS lightsail without env id and driver name was not correct.

image

royra:
fixes #159

add `listMachines` in drivers, in addition to `listDeletableResources`

`listMachines` is used by the `ls` command
`listDeletableResources` is used by the `purge` command

For drivers with only machine resources (all but AWS) the two functions are identical. Most of the diff is layout changes due to extracting `listMachines` and using it in two places.

## This is a

- [X] Bug fix
- [ ] Feature
- [ ] Doc improvement

- [X] I tested the proposed change on my cloud provider: AWS

### DIFF
--- a/packages/core/src/commands/ls.ts
+++ b/packages/core/src/commands/ls.ts
@@ -6,6 +6,6 @@ const ls = async ({
 }: {
   machineDriver: MachineDriver
   log: Logger
-}) => machineDriver.listDeletableResources()
+}) => machineDriver.listMachines()
 
 export default ls
--- a/packages/core/src/driver/driver.ts
+++ b/packages/core/src/driver/driver.ts
@@ -40,6 +40,7 @@ export type MachineDriver<
     stdio: PartialStdioOptions,
   ) => Promise<{ code: number } | { signal: string }>
 
+  listMachines: () => AsyncIterableIterator<Machine | PartialMachine>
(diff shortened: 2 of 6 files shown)


Yshayy:
Maybe the purge abstraction should be in the level of the provider?
From a technical point, it might be different logic for deleting a specific environment than deleting all resources of Preevy (including all the environments).
In this case, the contract in the driver will be deleteEnvironment and purge/removePreevy.

For example (although it's not the current implementation), k8s driver could create a namespace for all preevy assets and delete the namespace when purging. (instead of deleting them one by one).

royra:
> Maybe the purge abstraction should be in the level of the provider? From a technical point, it might be different logic for deleting a specific environment than deleting all resources of Preevy (including all the environments). In this case, the contract in the driver will be deleteEnvironment and purge/removePreevy.

Agree.

The purge flow is a bit more involved: presenting the resources to be deleted by type, get confirmation and then delete. It also accepts an optional `--all` flag to delete static resources (e.g, AWS keypairs). 

So the contract would have to be something like 
- `listPurgeResources` - returns the list, does not delete
- `purgeResources(includeStatic: boolean)` - does the delete

I can try implementing it in the current PR

PR #171, opened 2023-08-08
Refactor environment ID usage

royra:
- Tunnel server now accepts a `tunnel-url` command over ssh
- `up` and `urls` command now use the `tunnel-url` command instead of building the URLs themselves
- Tunnel server now normalizes the received forward requests to be DNS-safe
- Ambient Environment ID now includes only safe characters and max length
- Added more printed messages to communicate progress

## This is a

- [X] Feature

- [X] I tested the proposed change on my cloud provider: AWS


Yshayy:
I think we should split the logic of the tunnel-server and using tunnel-url function (for env variables) from "Ambient Environment ID" (and envid) logic changes and usage in urls.

I'm not sure what are the limitations we should set in that regard, maybe it's worth using the same conventions as docker-compose project names. (and maybe consider using the project name as -p for the environment to remove this additional configuration complexity)

In addition, I think there might be regression there as we currently allow deleting (down) env without the docker-compose (model evaluation is done only when --id is not specified), same for urls (as it can be based on the compose-tunnel-agent endpoint), related to #114.

The logic for tunnel-url makes sense for injecting variables, I believe most functionality of preevy should be able to perform without the local compose files (similar to docker compose -p).
In the future, it might be worth saving the compose file in the machine and use the remove model if it's not available locally, but for now I think it worth further discussion.

royra:
> I think we should split the logic of the tunnel-server and using tunnel-url function (for env variables) from "Ambient Environment ID" (and envid) logic changes and usage in urls.
> 
> I'm not sure what are the limitations we should set in that regard, maybe it's worth using the same conventions as docker-compose project names. (and maybe consider using the project name as -p for the environment to remove this additional configuration complexity)
> 
> In addition, I think there might be regression there as we currently allow deleting (down) env without the docker-compose (model evaluation is done only when --id is not specified), same for urls (as it can be based on the compose-tunnel-agent endpoint), related to #114.
> 
> The logic for tunnel-url makes sense for injecting variables, I believe most functionality of preevy should be able to perform without the local compose files (similar to docker compose -p). In the future, it might be worth saving the compose file in the machine and use the remove model if it's not available locally, but for now I think it worth further discussion.

Compose files are only required because of plugin definition.
If you specify `--id` there is no need for a compose file (except for the `up` command which inherently uses them of course)

As for the rest of the comments, let's discuss f2f.

PR #174, opened 2023-08-09
more metadata

royra:
- allow CTA to load metadata from multiple files in addition to the env var
- add driver metadata
- add creation date and version metadata

## This is a

- [X] Feature

- [X] I tested the proposed change on my cloud provider: GCE

### DIFF
--- a/packages/compose-tunnel-agent/src/metadata.ts
+++ b/packages/compose-tunnel-agent/src/metadata.ts
@@ -1,0 +1,34 @@
+import fs from 'fs'
+import { Logger } from '@preevy/common'
+import { merge } from 'lodash'
+import { inspect } from 'util'
+
+export const envMetadata = async ({ env, log }: { env: NodeJS.ProcessEnv; log: Logger }) => {
+  const jsons = [
+    ...await Promise.all(
+      (env.ENV_METADATA_FILES || '')
+        .split(' ')
+        .map(async f => {
+          try {
+            return await fs.promises.readFile(f, { encoding: 'utf8' })
+          } catch (err) {
+            log.warn('error reading env metadata from file "%s": %j', f, inspect(err))
+            return undefined
+          }
+        })
+    ),
+    env.ENV_METADATA,
+  ]
+
+  const objects = jsons.map((s, i) => {
+    try {
+      if (s === undefined) return undefined
+      return JSON.parse(s) as Record<string, unknown>
+    } catch (err) {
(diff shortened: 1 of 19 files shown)


packages/compose-tunnel-agent/src/metadata.ts
```diff
@@ -0,0 +1,34 @@
+import fs from 'fs'
+import { Logger } from '@preevy/common'
+import { merge } from 'lodash'
+import { inspect } from 'util'
+
+export const envMetadata = async ({ env, log }: { env: NodeJS.ProcessEnv; log: Logger }) => {
+  const jsons = [
+    ...await Promise.all(
+      (env.ENV_METADATA_FILES || '')
```
Yshayy, line 9: ENV_METADATA_FILES might be a bit of a problem if the folder contains 'space' because of the split. 
I think it might worth using a ENV_METADATA_DIR: 
can contain git metadata as well, but it can scan all the json files in the folder.

Alternatively, we can use a single merged json with sections. (can be passed either as a single file, or json var)
(in this case, it would be ENV_METADATA_FILE or ENV_METADATA)

Yshayy: Also, we can support both metadata injections mechanisms as it can make it more user-friendly to consume and test, but for preevy cli it might make it simpler to use the same mechanism for git and cloud metadata.
I also we have the joined type of "EnvMetadata", but I'm not sure where we use this.

Yshayy (approved):
Looks good overall,
Added a few notes and suggested minor changes.

packages/compose-tunnel-agent/src/metadata.ts
royra: Re pre-merged JSON suggestion: The driver metadata is generated at machine creation and saved on the machine persistent storage, so merging it at the `up` command would require reading it from the remote machine.

Re the directory suggestion: merge order is significant. I considered using a directory, but the order is not guaranteed. We can sort the directory listing and use filenames like `00-git.json`, `10-driver.json` but that requires knowing the order in advance, when generating the contents, or re-ordering the files JIT, again requiring reading the file from the remote machine.

Re spaces in filenames: I'll add an option to use a JSON array in the file list var, should solve this.

royra: > we have the joined type of "EnvMetadata", but I'm not sure where we use this.

It's the output of the `envMetadata` function, mostly to ensure the expected schema.

Issue #34, opened 2023-03-23
global flags not shown on --help


royra:
`preevy --help` should show all global flags (e.g, `--debug`)

Abhi5Pande:
Can I work on this?

sameerkali:
can someone explain what's going on here

royra:
Assigning @Abhi5Pande since they were first.

Oclif's notion of "global flags" [attaches flags to specific commands](https://oclif.io/docs/global_flags), so this may prove trickier than it seems. If you need to make compatibility breaking-changes LMK. You're invited discuss this further [on our Slack](https://join.slack.com/t/preevyusers/shared_invite/zt-1shi7t2dy-~glGUC5pWCtGrH19bKNXrQ).

royra:
@Abhi5Pande any progress?

Abhi5Pande:
Yes @royra , experimenting with custom help class and figuring out how we can get the list of global flags

PR #177, opened 2023-08-11
Added custom help logic

Abhi5Pande:
## This is a

- [ ] Bug fix - https://github.com/livecycle/preevy/issues/34

### DIFF
--- a/packages/cli/src/help/flag.ts
+++ b/packages/cli/src/help/flag.ts
@@ -1,0 +1,78 @@
+import chalk from 'chalk'
+import { Command } from '@oclif/core'
+import { HelpFormatter } from '@oclif/core/lib/help/formatter'
+import { BaseCommand } from '@preevy/cli-common'
+
+const {
+  underline,
+} = chalk
+const {
+  dim,
+} = chalk
+
+export default class Flag extends HelpFormatter {
+  public getFlags():string {
+    const flags = Object.entries(BaseCommand.baseFlags)
+      .filter(([_name, flag]) => !flag.hidden)
+      .map(([name, flag]) => {
+        flag.name = name
+        return flag
+      })
+
+    const body = this.flagFormatter(flags)
+
+    return this.section('GLOBAL FLAGS', body)
+  }
+
+  protected flagHelpLabel(flag: Command.Flag.Any, showOptions = false): string {
+    let label = flag.helpLabel
+
+    if (!label) {
+      const labels = []
+      if (flag.char) labels.push(`-${flag.char[0]}`)
+      if (flag.name) {
+        if (flag.type === 'boolean' && flag.allowNo) {
+          labels.push(`--[no-]${flag.name.trim()}`)
+        } else {
+          labels.push(`--${flag.name.trim()}`)
+        }
+      }
+
+      label = labels.join(', ')
+    }
+
+    if (flag.type === 'option') {
+      let value = flag.helpValue || (this.opts.showFlagNameInTitle ? flag.name : '<value>')
+      if (!flag.helpValue && flag.options) {
+        value = showOptions || this.opts.showFlagOptionsInTitle ? `${flag.options.join('|')}` : '<option>'
+      }
+
+      if (flag.multiple) value += '...'
+      if (!value.includes('|')) value = underline(value)
+      label += `=${value}`
+    }
+
+    return label
+  }
+
+  protected flagFormatter(flags: Array<Command.Flag.Any>): [string, string | undefined][] | undefined {
+    if (flags.length === 0) return undefined
+
+    return flags.map(flag => {
+      const left = this.flagHelpLabel(flag)
+
+      let right = flag.summary || flag.description || ''
+      if (flag.type === 'option' && flag.default) {
+        right = `[default: ${flag.default}] ${right}`
+      }
+
+      if (flag.required) right = `(required) ${right}`
+
+      if (flag.type === 'option' && flag.options && !flag.helpValue && !this.opts.showFlagOptionsInTitle) {
+        right += `\n<options: ${flag.options.join('|')}>`
+      }
+
+      return [left, dim(right.trim())]
+    })
+  }
+}
--- a/packages/cli/src/help/index.ts
+++ b/packages/cli/src/help/index.ts
@@ -1,28 +1,11 @@
 import { Help } from '@oclif/core'
-import { BaseCommand } from '@preevy/cli-common'
-import os from 'os'
+import Flag from './flag'
+// import FlagFormatter from './flag'
 
 export default class CustomHelpClass extends Help {
   protected async showRootHelp(): Promise<void> {
     await super.showRootHelp()
-    this.log(this.formatFlag())
-  }
-
-  protected formatFlag():string {
- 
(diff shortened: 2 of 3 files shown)


packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,72 @@
+import { Help } from '@oclif/core'
```
royra, line 1: I assume some of this code is taken from the base class. If I'm correct, it's better to reuse that code (e.g, inherit) than to copy it.

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    const flags: Flag[] = []
+
+    for (const flagName of Object.keys(BaseCommand.baseFlags)) {
+      if (BaseCommand.baseFlags.hasOwnProperty.call(BaseCommand.baseFlags, flagName)) {
+        const flagKey = flagName as keyof typeof BaseCommand.baseFlags
+        const flag = BaseCommand.baseFlags[flagKey]
+        if (flag.helpGroup === 'GLOBAL') {
+          const flagHelp: Flag = {
+            name: flagName,
+            description: flag.summary !== undefined ? flag.summary : '',
```
royra, line 55:
```suggestion
            description: flag.summary ?? '',
```

royra (commented):
Thanks for you contribution!

packages/cli/src/help/index.ts
Abhi5Pande: Hey yes you are correct, some part of it is pasted from the base class itself.
So should I use super or something to achieve that?

Abhi5Pande:
Hey! I have made the changes proposed by you. Please review... If there is anything more you would like different lmk.

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  protected formatFlag():string {
+    type Flag = {
+      name: string
+      description: string
+    };
+
+    const flags: Flag[] = []
+
+    for (const flagName of Object.keys(BaseCommand.baseFlags)) {
+      if (BaseCommand.baseFlags.hasOwnProperty.call(BaseCommand.baseFlags, flagName)) {
```
royra, line 19: why is this filter needed? there is no prototypical inheritance in `baseFlags` and we don't need to support it.

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,42 @@
...
+        const flag = BaseCommand.baseFlags[flagKey]
+        if (flag.helpGroup === 'GLOBAL') {
+          const flagHelp: Flag = {
+            name: flagName,
+            description: flag.summary ?? '',
+          }
+          flags.push(flagHelp)
+        }
+      }
+    }
```
royra, line 30: this can be simplified to:

```suggestion
    const flags = Object.entries(BaseCommand.baseFlags)
      .filter(([_name, flag]) => flag.helpGroup === 'GLOBAL')
      .map(([name, flag]) => ({ name, description: flag.summary ?? '' }))
```

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    const body = this.renderList(flags.map(c => [
+      `--${c.name}`,
+      c.description,
+    ]), {
+      spacer: '\n',
```
royra, line 35: use `os.EOL` for windows compatibility

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,42 @@
+import { Help } from '@oclif/core'
+import { BaseCommand } from '@preevy/cli-common'
+
+export default class MyHelpClass extends Help {
```
royra, line 4: the name `MyHelpClass` doesn't reflect the functionality this class

packages/cli/src/help/index.ts
```diff
@@ -0,0 +1,42 @@
...
+          const flagHelp: Flag = {
+            name: flagName,
+            description: flag.summary ?? '',
+          }
+          flags.push(flagHelp)
+        }
+      }
+    }
+    const body = this.renderList(flags.map(c => [
+      `--${c.name}`,
```
royra, line 32: There is probably rendering logic implemented in oclif that needs to be reused here.

For example,
- flags with short syntax (--debug, -D)
- boolean flags with the "no-" prefix

packages/cli/src/help/index.ts
Abhi5Pande: How about customHelpClass ?

packages/cli/src/help/index.ts
Abhi5Pande: Hey @royra, So the rendering is in another class in oclif and it is a protected function. The process of creating an instance of that class demands certain constructor parameters that hold no relevance for our current use case. In this case would it be okay if I copy pasted the logic of formatting instead of inheriting? Is there any better way to do it?

royra: Which protected function are you referring to?

Abhi5Pande: So there is another class in oclif called CommandHelp, which is basically the logic for command help. Oclif shows global flags when --help flag is added to a command and the logic of formatting the flags is in that class itself. 

We essentially need to use the protected flags function:

      protected flags(flags: Array<Command.Flag.Any>)

This function returns the body string in the format we need which we can then render as we are doing already

royra: You could inherit from `CommandHelp` and expose it.

Abhi5Pande:
@royra  Can you review and share your thoughts on this

packages/cli/src/help/flag.ts
```diff
@@ -0,0 +1,78 @@
...
+import { Command } from '@oclif/core'
+import { HelpFormatter } from '@oclif/core/lib/help/formatter'
+import { BaseCommand } from '@preevy/cli-common'
+
+const {
+  underline,
+} = chalk
+const {
+  dim,
+} = chalk
```
royra, line 11:
```suggestion
import { underline, dim } from 'chalk'
import { Command } from '@oclif/core'
import { HelpFormatter } from '@oclif/core/lib/help/formatter'
import { BaseCommand } from '@preevy/cli-common'

```

packages/cli/src/help/flag.ts
```diff
@@ -0,0 +1,78 @@
...
+        flag.name = name
+        return flag
+      })
+
+    const body = this.flagFormatter(flags)
+
+    return this.section('GLOBAL FLAGS', body)
+  }
+
+  protected flagHelpLabel(flag: Command.Flag.Any, showOptions = false): string {
```
royra, line 27: instead of copying the oclif code (which can change and break stuff), it's possible to inherit from the original `CommandHelp` to modify it and use it in the custom `Help` class. There is an example at the oclif tests: https://github.com/oclif/core/blob/c992403175c40492a5b724af3c117c6a059b4843/test/help/show-customized-help.test.ts#L14

royra (commented):
Hi @Abhi5Pande! Sorry for the late review.
See comments below.
Also, looks like this works with `preevy --help` but not with `preevy help` - the latter does not print the "global flags" section.

royra:
issue fixed by #306

PR #184, opened 2023-08-14
Add auth scheme hint query param

AssafKr:
## This is a

- [ ] Bug fix
- [ ] Feature
- [ ] Doc improvement

### DIFF
--- a/packages/core/src/compose-tunnel-agent-client.ts
+++ b/packages/core/src/compose-tunnel-agent-client.ts
@@ -175,7 +175,12 @@ export const queryTunnels = async ({
       .filter(({ service }: Tunnel) => showPreevyService || service !== COMPOSE_TUNNEL_AGENT_SERVICE_NAME)
       .map(tunnel => ({
         ...tunnel,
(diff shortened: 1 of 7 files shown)


Yshayy (approved):
LGTM :)

packages/core/src/compose-tunnel-agent-client.ts
```diff
@@ -175,7 +178,12 @@ export const queryTunnels = async ({
       .filter(({ service }: Tunnel) => showPreevyService || service !== COMPOSE_TUNNEL_AGENT_SERVICE_NAME)
       .map(tunnel => ({
         ...tunnel,
-        ports: mapValues(tunnel.ports, includeAccessCredentials ? addCredentials : (x: string) => x),
+        ports: mapValues(
+          tunnel.ports,
+          includeAccessCredentials
+            ? (x: string) => `${addCredentials(x)}?_preevy_auth_hint=basic`
```
royra, line 184: Adding the credentials using string concatenation might break if the URL already contains query parameters.

I suggest creating `setQueryParam` alongside `replaceHostname` at `common/src/urls.ts`

PR #188, opened 2023-08-16
gce: fix labels

### LINKED ISSUE #153: Preevy deployment with Google cloud
Currently, if you'll try to deploy with the GCP driver while your branch name contains Uppercase characters, you'll get the following error

```
Run echo "/home/runner/work/_actions/livecycle/preevy-up-action/v1.2.0" >> $GITHUB_PATH
Run # Convert the docker-compose-yaml-paths input to an env var
Docker file paths docker-compose.yml
Running Preevy with files: -f docker-compose.yml -f docker-compose.dev.yml 
Using environment ID: preevy-saas-eliyaenvironmentsUI
- Creating Google Cloud machine
✖ Creating Google Cloud machine: Creating from existing snapshot
    Error: Invalid value for field 'resource.labels': ''. Label value 
    'preevy-saas-eliyaenvironmentsUI' violates format constraints. The value 
    can only contain lowercase letters, numeric characters, underscores and 
    dashes. The value can be at most 63 characters long. International 
    characters are allowed.
    Code: 3
```

Possible solution: lowercase the resource label value.

royra:
label contained the envId as is, which does not adhere to the [allowed characters](https://cloud.google.com/compute/docs/labeling-resources#requirements).

fixed by storing a hash of the envId and profileId, in the labels, which are used as a filter when querying the instances. actual values are stored in the [instance metadata](https://cloud.google.com/compute/docs/metadata/setting-custom-metadata) and returned for the ls command.

fixes #153

### DIFF
--- a/packages/driver-gce/src/driver/client.ts
+++ b/packages/driver-gce/src/driver/client.ts
@@ -2,7 +2,7 @@ import { InstancesClient, ImagesClient, ZoneOperationsClient, RegionsClient } fr
 import { GoogleError, Status, operationsProtos, CallOptions } from 'google-gax'
 import { asyncFirst } from 'iter-tools-es'
 import { randomBytes } from 'crypto'
-import { LABELS, sha1hex } from './labels'
+import { LABELS, isValidLabel, normalizeLabel } from './labels'
 import { readCloudConfig } from '../static'
 import { metadataKey, serializeMetadata } from './metadata'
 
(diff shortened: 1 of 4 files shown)


Yshayy (approved):

Yshayy:
Will preevy down or update work for existing environments?

royra:
> Will preevy down or update work for existing environments?

No. To do that I need to change the filter to include the old labels as well as the new for listing. For that I'll also need to implement a check that the specified env id is a valid label (otherwise it will throw). didn't think it's worth complicating the code

PR #195, opened 2023-08-21
Feature - support tunneling existing compose app

### LINKED ISSUE #69: Support specifying which services to expose from the environments
Currently, preevy expose all services by default

Yshayy:
## This is a
- [x] Bug fix
- [x] Feature
- [ ] Doc improvement

Fixes #69, update behavior exposed tunnels in Preevy cta/agent:
- Respecting by default docker port binding (by default - expose only if publicPort)
- Support specifying explicit port expose by using (preey.expose=port0,port1)
- If using `preevy.expose` label, only ports that are defined in `preevy.expose` AND have either definition in the docker image (expose command) or specified in the compose file will expose. (this behavior might change in the future) 

Bug fix - `--json` flag not respected in the output of profile ls/current commands

Feature:
Support new (hidden) cli proxy commands to tunnel existing running compose application.
`preevy proxy connect <compose-project>` - attach preevy agent to existing compose application and produce tunneled URLs for running services.
`preevy proxy disconnect <compose-project>` - remove preevy agent from existing compose application effectively stop tunneling
`preevy proxy urls <compose-project>` - query attached preevy agent for urls, support all the regular flags of `preevy urls` command


packages/cli/src/commands/proxy/connect.ts
```diff
@@ -0,0 +1,127 @@
...
+
+// eslint-disable-next-line no-use-before-define
+export default class Connect extends ProfileCommand<typeof Connect> {
+  static description = 'Tunnel existing local compose application'
+
+  static flags = {
+    ...tunnelServerFlags,
+    ...urlFlags,
+    ...ux.table.flags(),
+    'private-env': Flags.boolean({
```
AssafKr, line 20: I think private-env isn't that clear. Maybe something like "--access=private", WDYT?

packages/cli/src/commands/proxy/connect.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  static args = {
+    'compose-project': Args.string({
+      description: 'Compose Project name',
+      required: true,
+    }),
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async run(): Promise<unknown> {
```
AssafKr, line 38: If I'm looking at other command, the UI/UX stuff is in the cli/src/commands folder and the business logic is in the core/src folder, consider moving the logic there

packages/cli/src/commands/proxy/connect.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const tunnelingKey = await pStore.getTunnelingKey()
+
+    const tunnelOpts = {
+      url: flags['tunnel-url'],
+      tlsServerName: flags['tls-hostname'],
+      insecureSkipVerify: flags['insecure-skip-verify'],
+    }
+
+    const envId = `local-${Math.random().toString(36).substring(7)}` // local+find_ambient_id_based on compose dir (?)
```
AssafKr, line 51: why the "local" prefix?

AssafKr (approved):

packages/cli/src/commands/proxy/connect.ts
Yshayy: Since it's a local environment.
I'll improve it once we've read docker-compose-dir label

PR #198, opened 2023-08-24
Minor fixes in tunnel server

Yshayy:
## This is a
- [x] Bug fix

Small change to the logic to generate active tunnel key store (+tests).
Add warning and conditionals for debugging/handling the cases of duplicate path errors.


tunnel-server/src/strings.ts
```diff
@@ -1,12 +1,13 @@
...
 
-export const truncateWithHash = (s: string, maxLength: number, hashLength = 4, sep = '-') => {
-  if (s.length <= maxLength) {
+export const truncateWithHash = (s: string, sanitize: (s:string)=> string, maxLength: number, hashLength = 4, sep = '-') => {
+  const sanitized = sanitize(s)
+  if (s.length <= maxLength && sanitized === s) {
     return s
   }
 
   const suffix = createHash('md5').update(s).digest('base64url').replace(/[^a-zA-Z0-9]/g, '')
```
royra, line 9: looks like `sanitize` needs to run on the suffix as well

tunnel-server/src/tunnel-store.ts
```diff
@@ -54,3 +55,19 @@ export const inMemoryActiveTunnelStore = ({ log }: { log: Logger }): ActiveTunne
...
   }
 }
+
+const MAX_DNS_LABEL_LENGTH = 63
+
+export const activeTunnelStoreKey = (clientId: string, remotePath: string) => {
+  // return value needs to be DNS safe:
+  // - max DNS label name length is 63 octets (== 63 ASCII chars)
+  // - case insensitive
+  const tunnelPath = remotePath.substring(1)
```
royra, line 65: this removes the leading `/`? if so, it's duplicated below by `x.replace(/^\//, '')`
IMO the latter is better since it clarifies the purpose.

royra (commented):
see comments below

tunnel-server/src/tunnel-store.ts
Yshayy: fixed

tunnel-server/src/strings.ts
Yshayy: Added check on the client-id sanitization + test

tunnel-server/src/tunnel-store.ts
```diff
@@ -54,3 +55,24 @@ export const inMemoryActiveTunnelStore = ({ log }: { log: Logger }): ActiveTunne
...
   }
 }
+
+const MAX_DNS_LABEL_LENGTH = 63
+
+const sanitizeHostName = (x:string) => x.replace(/[^a-zA-Z0-9_-]/g, '-').toLocaleLowerCase()
+export const activeTunnelStoreKey = (clientId: string, remotePath: string) => {
+  // return value needs to be DNS safe:
+  // - max DNS label name length is 63 octets (== 63 ASCII chars)
+  // - case insensitive
```
royra, line 65: this comment now looks out of place

Yshayy: Fixed, add JSDoc instead

royra (approved):

PR #201, opened 2023-08-27
add profile:key command to show key

royra:
command can show any of the following
- public key (in PEM or SSH formats)
- private key
- thumbprint (in short or full URI formats)

### DIFF
--- a/packages/cli/src/commands/profile/key.ts
+++ b/packages/cli/src/commands/profile/key.ts
@@ -6,12 +6,12 @@ import ProfileCommand from '../../profile-command'
 const keyTypes = ['private', 'public-pem', 'public-ssh', 'thumbprint', 'thumbprint-uri'] as const
 type KeyType = typeof keyTypes[number]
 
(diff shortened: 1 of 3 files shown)


Yshayy (approved):
LGTM :)

packages/cli/src/commands/profile/key.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (type === 'private') {
+    return parseKey(key).getPrivatePEM()
+  }
+
+  throw new Error(`Invalid key type "${type}"`)
+}
+
+// eslint-disable-next-line no-use-before-define
+export default class CurrentProfile extends ProfileCommand<typeof CurrentProfile> {
```
Yshayy, line 30: Change the name of the class

packages/cli/src/commands/profile/key.ts
```diff
@@ -0,0 +1,56 @@
...
+  static description = 'Display current profile in use'
```
Yshayy, line 31: Change description

PR #204, opened 2023-08-28
some minor fixes

royra:
- renamed metadata property `driver` to `machine`
- gce driver: fixed error when running up while previous env machine is being deleted
- upgrade async-iter-es to latest version


Yshayy (approved):
LGTM overall :)
I'd consider keeping the previous names of getEnvMachine/getEnvMachineAndSpecDiff. (the signature already explicitly specify envId as param)

PR #206, opened 2023-08-28
Allow specify envId in connect command

Yshayy:
Add tagged type for envId.
Unified envId behavior in commands.up ->
Fix a missing normalization bug of envId in up command.

## This is a

- [x] Bug fix
- [x] Feature
- [ ] Doc improvement


packages/cli/src/commands/proxy/connect.ts
```diff
@@ -49,7 +53,7 @@ export default class Connect extends ProfileCommand<typeof Connect> {
     }
     const composeProject = args['compose-project']
     const deviceId = (await machineId(this.config.dataDir)).substring(0, 2)
-    const envId = normalizeEnvId(`${composeProject}-dev-${deviceId}`) // local+find_ambient_id_based on compose dir (?)
+    const envId = flags['env-id'] ? validateEnvId(flags['env-id']) : normalizeEnvId(`${composeProject}-dev-${deviceId}`)
```
royra, line 56: I would print a log describing how the id got created (if not specified explicitly), like the other commands do.

packages/cli/src/commands/proxy/connect.ts
```diff
@@ -16,6 +16,10 @@ export default class Connect extends ProfileCommand<typeof Connect> {
     ...tunnelServerFlags,
     ...urlFlags,
     ...ux.table.flags(),
+    'env-id': Flags.string({
+      description: 'specify environment id for this app',
+      required: false,
+    }),
```
royra, line 22: Not a part of this commit, bit perhaps `--access private` is better than the boolean flag

royra (approved):

packages/core/src/env-id.ts
```diff
@@ -1,12 +1,16 @@
...
 
+export type EnvId = string & {
+  __tag: 'EnvId'
+}
+
 export const normalize = (s: string) => s
   .toLowerCase()
   .replace(/[^a-z0-9_-]/g, '-')
-  .replace(/^[^a-z]/, firstChar => `a${firstChar}`) // prepend alpha char if first char is not alpha
+  .replace(/^[^a-z]/, firstChar => `a${firstChar}`) as EnvId
```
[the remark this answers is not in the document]
Yshayy: Nope, adding back

packages/core/src/env-id.ts
```diff
@@ -1,12 +1,16 @@
 import { detectCiProvider } from './ci-providers'
-import { gitBranchName } from './git'
+import { gitContext } from './git'
 import { ComposeModel } from './compose'
 import { Logger } from './log'
 
+export type EnvId = string & {
+  __tag: 'EnvId'
```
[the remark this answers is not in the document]
Yshayy: I've originally used it in connect to make sure that I'm passing envId, but it detected a bug in commands.up where flag.id was used without normalization. (also fixed in part of this PR)
I currently use it for all initial input validation and building the compose model. 

It's possible to add it to the driver signature as well, but the driver phase is pretty far from the CLI input and env detection that I believe it doesn't add much value.

packages/cli/src/commands/proxy/connect.ts
Yshayy: I think it's a bit tricky to communicate concisely.
The current implementation is a mix of compose-project, dev suffix and deviceId.

I can use `generated` vs `user specified`, but I'm not sure if it reveals useful information as the user know they specify the id.

packages/cli/src/commands/proxy/connect.ts
Yshayy: I'll fix it in another PR, but I'm not sure what's the right flag here, so it can be done in another PR. (the command is still hidden)
In practice, it defines "default-access" (since it can be overridden by labels), but I think that `--access` or `--default-access` might be ambiguous.

packages/cli/src/commands/proxy/connect.ts
```diff
@@ -16,6 +16,10 @@ export default class Connect extends ProfileCommand<typeof Connect> {
     ...tunnelServerFlags,
     ...urlFlags,
     ...ux.table.flags(),
+    'env-id': Flags.string({
```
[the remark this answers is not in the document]
Yshayy: Changed to id.
Added `env-id` as flag alias, it might be worth adding the alias to other commands and in the future switch between the alias and the flag as ID can be ambiguous

packages/cli/src/commands/proxy/connect.ts
Yshayy: Added log based on Docker Machine and `Local device id`

PR #209, opened 2023-08-30
cta: narrow down docker API

royra:
replace generic docker API proxy with specific endpoints


packages/compose-tunnel-agent/src/api-server/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  log: Logger
+  currentSshState: () => Promise<SshState>
+  machineStatus?: () => Promise<{ data: Buffer; contentType: string }>
+  envMetadata?: Record<string, unknown>
+  composeModelPath: string
+  dockerFilter: DockerFilterClient
+  docker: Dockerode
+}) => {
+  const handler = tryHandler({ log }, async (req, res) => {
+    const { pathname: path } = url.parse(req.url || '')
```
Yshayy, line 32: Need to take more careful look, but I think that since this api is know behaving very similar to a regular rest-api, it might worth using express/fastify or other package to reduce the noise and high-level implementations of routing.

packages/compose-tunnel-agent/src/api-server/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    if (req.method === 'GET' && path === '/healthz') {
```
Yshayy, line 34: Not specifically for this PR, but teed to add support for CORS, so these apis can be fetched also from client.
@eliyabar currently add it locally for testing

packages/compose-tunnel-agent/src/api-server/ws/handlers/exec.ts
```diff
@@ -0,0 +1,62 @@
...
+import { NotFoundError } from '../../http-server-helpers'
+
+const handler = wsHandler(
+  /^\/container\/([^/?]+)\/exec($|\?)/,
+  async (ws, req, match, { log, docker, dockerFilter }) => {
+    const containerId = match[1]
+    if (!await dockerFilter.inspectContainer(containerId)) {
+      throw new NotFoundError()
+    }
+    const { obj: { tty: ttyQueryParam }, search } = parseQueryParams(req.url ?? '', { tty: true })
```
Yshayy, line 14: It might worth splitting it to two ops in the future:
One for creating the exec/logs handler (return handle url), the other for reading in websocket based on the handle.
The handle can up-for-grabs and available in a limited time window. (can be saved even in CTA memory, or localfs)

It can be a bit safer as POST requests can work with authorization header and CORS.

packages/core/src/compose-tunnel-agent-client.ts
```diff
@@ -70,6 +72,10 @@ export const addComposeTunnelAgentService = (
     machineStatusCommand?: MachineStatusCommand
     envMetadata: EnvMetadata
     composeModelPath: string
+    composeProject: string
+    profileThumbprint?: string
+    privateMode: boolean
```
Yshayy, line 77: The privateMode can be derived from the defaultAccess paramter (or the opposite), it's a label with product semantics, while defaultAccess is a CTA env var, but they are basically the same. (The label is currently only read by the Docker extension AFAIK)

packages/compose-tunnel-agent/src/api-server/index.ts
royra: converted to fastify

packages/compose-tunnel-agent/src/api-server/index.ts
royra: done

packages/core/src/compose-tunnel-agent-client.ts
royra: can the `privateMode` be read from the `/env-metadata` endpoint instead of the label?

PR #220, opened 2023-09-10
tunnel server: safer delete

royra:
- do not delete entries from other clients
- remove store entry on tcp server error


tunnel-server/src/tunnel-store/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      if (keyToTunnel.has(key)) {
+        throw new KeyAlreadyExistsError(key)
+      }
+      const txId = txIdGen.next()
+      log.debug('setting tunnel %s: %j', key, txId, value)
+      keyToTunnel.set(key, Object.assign(value, { txId }))
+      pkThumbprintToTunnel.add(value.publicKeyThumbprint, value)
+      return { txId }
+    },
+    delete: async (key, tx) => {
```
Yshayy, line 71: Maybe it's worth returning value if the delete was not commited

Yshayy (approved):
LGTM 👍 
Added few notes.
In general, I think we should try to simplify it a bit after it's stabilized.

tunnel-server/src/tunnel-store/array-map.ts
```diff
@@ -0,0 +1,33 @@
+export type ArrayMap<K, V> = {
+  get: (key: K) => readonly V[] | undefined
+  add: (key: K, value: V) => void
+  delete: (key: K, pred: (value: V) => boolean) => void
+}
+
+export const arrayMap = <K, V>(): ArrayMap<K, V> => {
```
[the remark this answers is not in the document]
royra: 👍🏽

tunnel-server/src/tunnel-store/index.ts
royra: in a generic implementation, absolutely. I just don't see a use for a return value for the specific use case we have, so I think adding it (and testing for it) is not worth it now.

Yshayy: It's not a must, but if we want to add visibility/log that the row was indeed deleted after being called from the client. (due to the issue)

PR #227, opened 2023-09-18
tunnel server: check and close inactive connections

royra:


tunnel-server/src/ssh/index.ts
```diff
@@ -22,48 +22,62 @@ export const createSshServer = ({
...
+          if (!(e instanceof KeyAlreadyExistsError)) {
+            throw e
+          }
+          const existingEntry = await activeTunnelStore.get(key)
+          if (!existingEntry) {
+            return await set() // retry
+          }
+          const otherClient = existingEntry.value.client as BaseSshClient
+          if (otherClient.connectionId === connectionId) {
+            throw new Error(`duplicate path: ${key}, from same connection ${connectionId}`)
```
Yshayy, line 57: Does the handling need to be different in this case?

```
if (otherClient.connectionId === connectionId) {
            throw new Error(`duplicate path: ${key}, from same connection ${connectionId}`)
          }
```
If we log it and continue with the same flow, will we get to a correct end state?

tunnel-server/src/ssh/index.ts
```diff
@@ -22,48 +22,62 @@ export const createSshServer = ({
...
+          }
+          if (!await otherClient.ping(5000)) {
+            const existingDelete = onceWithTimeout(existingEntry.watcher, 'delete', { milliseconds: 2000 })
+            void otherClient.end()
```
Yshayy, line 61: Do we need to handle the timeout here? If I follow the flow correctly, the request will be aborted in this case.

tunnel-server/src/memory-store.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (value && (setTx === undefined || value.setTx.txId === setTx.txId) && map.delete(key)) {
+        nextTick(() => { value.watcher.emit('delete') })
+        return true
+      }
+      return false
+    },
+  }
+}
+
+export type Store<V extends {}> = ReturnType<typeof inMemoryStore<V>>
```
Yshayy, line 53: Splitting the logic of the watcher to an external unit (a bit like a decorator style) might simple the code a bit.

So, 
```
store = new Store()
const storeWithEmitter = new StoreEmitter(store)
```

It can make the implementation also a bit easier to test.
Not sure if it's worth doing now...

Yshayy (approved):
LGTM :)
Added few notes.
I think that in the future, we need to see how we can simplify the implementation of tunnel server store (and the related units) as it combine event-driven, async and transaction semantics/mechanisms which can a bit complex to comprehend and verify correctness.

tunnel-server/src/ssh/index.ts
royra: a timeout means that ending the other client's connection did not result deleting the other client's entry from the store, and in this case the request should be aborted.

tunnel-server/src/ssh/index.ts
royra: requesting a duplicate forward in the same client connection should reject the request in the `baseSshServer`, so this is handled as an edge case which is essentially a bug.

PR #235, opened 2023-09-26
Added support for git context in proxy connect command

Yshayy:
Also Parametrized cwd for git utils.

## This is a

- [ ] Bug fix
- [x] Feature
- [ ] Doc improvement

- [x] I have used Preevy for a while and am familiar with the function it providers.


packages/core/src/commands/proxy.ts
```diff
@@ -30,36 +31,49 @@ export function inspectRunningComposeApp(projectName: string) {
...
-    const agentContainerId = await getPreevyAgentContainer()
-    if (agentContainerId) {
-      return await dockerCmd(`inspect ${agentContainerId}  --format '{{ index .Config.Labels "preevy.env_id"}}'`)
-    }
-    return null
+    const agentContainer = await getPreevyAgentContainer()
+    return agentContainer?.labels['preevy.env_id']
+  }
+
+  const listAllContainers = async () => ((await dockerCmd(`ps -a --filter ${projectFilter} --format json`)).split('\n') ?? [])
```
royra, line 53: at this point, isn't it better to use dockerode?

royra (approved):

packages/core/src/commands/proxy.ts
Yshayy: Yeah, I currently added this functionality, but when implementing it initially I use Dockerode.
There was a slight problem that it didn't understand docker contexts, by default it used:
unix:///var/run/docker.sock
But in docker for desktop, it changed in the last versions to:
unix:///{USER_HOME}/.docker/run/docker.sock
Which broke compatibility with the initial Dockerode defaults.
Back then, after playing it for a while and mange to make it work (reading config.json, and context files partially), I figured it's not worth the extra code and complexity at this point. (since we're already are using Docker cli)

PR #237, opened 2023-10-01
auth refactor

Yshayy:
## This is a

- [ ] Bug fix
- [ ] Feature
- [ ] Doc improvement

### DIFF
--- a/tunnel-server/src/auth.ts
+++ b/tunnel-server/src/auth.ts
@@ -44,7 +44,7 @@ export type BearerAuthorizationHeader = {
 }
 
 export type AuthorizationHeader = BasicAuthorizationHeader | BearerAuthorizationHeader
-export type JWTIssuer = {
+export type IdentityProvider = {
   issuer: string
   publicKey: KeyObject
   mapClaims: (issuer: JWTPayload, context: { pkThumbprint: string }) => Claims
@@ -86,7 +86,7 @@ const extractAuthorizationHeader = (req: IncomingMessage): AuthorizationHeader |
 
 export const jwtAuthenticator = (
   publicKeyThumbprint: string,
-  issuers: JWTIssuer[]
+  identityProviders: IdentityProvider[]
 ) : Authenticator => async req => {
   const authHeader = extractAuthorizationHeader(req)
   const jwt = match(authHeader)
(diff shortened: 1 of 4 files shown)


tunnel-server/src/auth.ts
```diff
@@ -42,15 +44,13 @@ export type BearerAuthorizationHeader = {
 }
 
 export type AuthorizationHeader = BasicAuthorizationHeader | BearerAuthorizationHeader
-
-const cookieName = 'preevy-saas-jwt'
-
-type VerificationData = {
-  pk: KeyObject
-  extractClaims: (token: JWTPayload, thumbprint: string) => Claims
+export type JWTIssuer = {
```
royra, line 47: the term `issuer` is used ambiguously in this flow for a name (string), object, payload - IMO need to replace with more specific terms.

tunnel-server/src/auth.ts
```diff
@@ -101,91 +101,72 @@ export const jwtAuthenticator = (
...
   if (parsedJwt.iss === undefined) throw new AuthError('Could not find issuer in JWT')
 
-  let verificationData: VerificationData
-  try {
-    verificationData = getVerificationData(parsedJwt.iss, publicKeyThumbprint)
-  } catch (e) {
-    if (e instanceof AuthError) return { isAuthenticated: false }
-
-    throw e
+  const jwtIssuer = issuers.find(x => x.issuer === parsedJwt.iss)
```
royra, line 104: probably should add a log here

tunnel-server/src/proxy/index.ts
```diff
@@ -164,7 +166,7 @@ export const proxy = ({
       const { req: mutatedReq, activeTunnel } = await validateProxyRequest(
         tunnelFinder,
         req,
-        pkThumbprint => sessionStore(req, undefined as unknown as ServerResponse, pkThumbprint),
+        pkThumbprint => sessionStore(req, new ServerResponse(req), pkThumbprint),
```
royra, line 169: not sure creating a throwaway response makes a clearer statement than passing `undefined`. I'll try to refactor this later so it'll be possible to create a readonly `cookieSessionStore` but better to leave it out of this PR due to possible conflicts.

royra (approved):
see minor comments

tunnel-server/src/auth.ts
Yshayy: I can use IdentityProvider

tunnel-server/src/proxy/index.ts
Yshayy: I started with using a throw-away mock proxy based object, it worked, but it based on the awareness of the cookie's internal details doesn't manipulate the returned object from the Server request.
I thought it was a bit safer and cleaner that way, but I can also create the throw-away object inside the session store so it won't leak.

tunnel-server/src/auth.ts
Yshayy: Currently it's the same behavior is before. (previous, it throw AuthError which is specially ignore)
Do we have a static logger here? If not, I'm not sure it's worth adding at this level. 
Since this method already throwing and having return value, it can delegate it externally without adding additional dependecy. (which can also be useful information for the app to decide whether to use 400/401)

tunnel-server/src/auth.ts
Yshayy: Updated it.

tunnel-server/src/proxy/index.ts
Yshayy: Move it inside the session store.
Does it conflict with your PR?

PR #293, opened 2023-10-18
preevy login at init

royra:
- some refactoring of cli
- add optional login to init command


packages/cli/src/commands/init.ts
```diff
@@ -28,137 +129,52 @@ export default class Init extends BaseCommand {
...
+      '--log-level', this.flags['log-level'] ?? 'info',
+      profileAlias,
+      location,
+      '--driver', driver,
+      ...formatDriverFlagsToArgs(driver, driverStatic.flags as Record<string, Flag<unknown>>, driverFlags),
+    ])
+
+    ux.info(text.recommendation('Use Livecycle together with Preevy to enable easy sharing and collaboration of your environments!'))
+
+    if (!await confirm({
```
Yshayy, line 168: I think we should emphasize to `new/existing Livecycle account` so it'll be clear that having an account is not a prerequisite. Alternatively, use something like `register profile on Livcycle`

PR #405, opened 2024-01-25
tunnel-server: add basic deployment guide

royra:

## This is a

- [ ] Bug fix
- [ ] Feature
- [X] Doc improvement

- [X] I have used Preevy for a while and am familiar with the function it provides.


tunnel-server/deployment/k8s/README.md
```diff
@@ -0,0 +1,100 @@
...
+Deploying a private instance of the Tunnel Server allows for fine-grained control of:
+
+- The URLs created for preview environments: e.g, use a custom domain.
+- Geolocation of the server: reduced distance to environments can result better network performance.
+- Security and privacy: deploy everything in your VPC, no traffic to 3rd parties.
+
+## Requirements
+
+- A Kubernetes cluster
+- An ingress solution to make K8S Services accesible from your network (e.g, Traefik). In this example, we'll use your cloud provider's load balancer.
```
Yshayy, line 18: The "An ingress solution" might be confusing here due to the ambiguity with k8s ingress resources. 
Since Traefik and cloud provider load balancer (based on k8s service-> loadbalancer) work at different layers (l4 vs l7), and the sentence imply that we're using the cloud provider's load balancer instead (which is incorrect, since we'll need service->loadbalancer endpoint to expose traefik if we were to use it).
In addition, cloud providers has both solutions for l4 and l7 integrated with Kubernetes (via k8s service->loadbalancer and k8s gateway/ingress) so it's not clear which one are we referring to.

tunnel-server/deployment/k8s/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Create two DNS records: `*.yourdomain.example` and `yourdomain.example`, both pointing to the external IP of the `tunnel-server` service.
+
+The address is not guaranteed to be static. According to your Kubernetes provider, there could be multiple ways to define a DNS entry for it. Here are some guides:
+
+- Amazon AWS: [EKS](https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html)
+- Google Cloud: [GKE](https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer)
+- Azure: [AKS](https://learn.microsoft.com/en-us/azure/aks/load-balancer-standard)
+
+Another approach would be to use a 3rd-party ingress solution like [Traefik](https://doc.traefik.io/traefik/user-guides/crd-acme/).
```
Yshayy, line 96: Traefik CRD won't solve the problem of configuring dns.
In k8s, there's external dns (https://github.com/kubernetes-sigs/external-dns) which works with many dns providers, but I didn't see wildcard support

Yshayy (commented):
Looks awesome, added few notes.

In the future, it might be worth having an example with k8s gateway APIs which means there's no need for stunnel/sslh.

TLSRoute can be used for the ssh traffic, and HTTPRoute for the HTTP traffic. (although in this case routing will be based on hostname and not on ALPN)

PR #454, opened 2024-03-25
CTA: fix dockerode filters

royra:


royra:
> Is it a type error or behavior error?

Both.

Yshayy (approved):

PR #479, opened 2024-04-21
urls: wait for all expected services

royra:
wait for all expected services to be returned from the CTA query


Yshayy (commented):
Will it work correctly for compose services that are ran to completion and not live (like init, data seeding, etc...)

royra:
Superseded by #488

PR #546, opened 2025-04-21
release 0.0.66

royra:
- **Bump the root-deps group across 1 directory with 37 updates**
- **fixes for dependabot version bumps**
- **Bump the tunnel-server-deps group across 1 directory with 17 updates**
- **fixes for depedabot version bumps (tunnel server)**
- **upgrade eslint**
- **fix oclif deps**


Yshayy (approved):
Do we need to deploy a new tunnel server as well?

royra:
> Do we need to deploy a new tunnel server as well?

Yep, dep updates

Yshayy:
Will the CLI work both against new tunnel server and old tunnel server?

royra:
> Will the CLI work both against new tunnel server and old tunnel server?
> 
> (I see also that most updates are dev deps)

Yes. There are no changes in behavior.

